PR #63, opened 2022-03-08
Hotfix 4.1.11

ZeroStride:
* Log out ISO8601 date time during sdk_init


Teak/TeakLog.m
```diff
@@ -113,7 +113,16 @@ - (nullable id)initForTeak:(nonnull Teak*)teak withAppId:(nonnull NSString*)appI
...
-  [self logEvent:@"sdk_init" level:INFO eventData:nil];
+
+  // Log ISO8601 format timestamp at init
+  NSDateFormatter* formatter = [[NSDateFormatter alloc] init];
+  [formatter setLocale:[NSLocale localeWithLocaleIdentifier:@"en_US_POSIX"]];
+  formatter.timeZone = [NSTimeZone timeZoneWithAbbreviation:@"UTC"];
+  [formatter setDateFormat:@"yyyy-MM-dd'T'HH:mm:ss"];
+  NSString* iso8601String = [formatter stringFromDate:[NSDate date]];
+  
+  NSDictionary* dateTime = @{@"at" : [iso8601String stringByAppendingString:@"Z"]};
```
AlexSc, line 124: Why not just put the 'Z' in as part of the date format?

Teak/TeakLog.m
```diff
@@ -113,7 +113,16 @@ - (nullable id)initForTeak:(nonnull Teak*)teak withAppId:(nonnull NSString*)appI
...
+  [self logEvent:@"sdk_init" level:INFO eventData:@{@"at" : dateTime}];
```
AlexSc, line 125: dateTime is already a dictionary, so this will give us `at.at`.

AlexSc (changes requested):

Teak/TeakLog.m
ZeroStride: It doesn't format that properly. It will be something like @"2022-03-08T01:49:57+0000" instead of @"2022-03-08T01:48:31Z"

AlexSc: The Z should be in '' like the T, so 'Z'

ZeroStride: Oh duh. Right.

Teak/TeakSession.m
```diff
@@ -277,7 +277,7 @@ - (void)sendUserIdentifier {
                                                       NSString* deepLink = reply[@"deep_link"];
                                                       NSURL* url = [NSURL URLWithString:deepLink];
                                                       if (url && blockSelf.launchDataOperation != nil) {
-                                                        [blockSelf.launchDataOperation.result updateDeepLink:url];
+                                                        blockSelf.launchDataOperation = [blockSelf.launchDataOperation updateDeepLink:url];
```
AlexSc, line 280: This will suppress the original short URL, whereas we preserve it on Android.

AlexSc (changes requested):

Teak/TeakSession.m
ZeroStride: Yep. There were some bigger changes needed to preserve it, and so I decided I'd do it in a future update.

ZeroStride: Changes made.

PR #90, opened 2025-10-10
[2025-10-17] Update build env to Xcode 14

AlexSc:
Given the challenges we're facing with teak-unity, let's keep updates minimal and done one at a time. This gets us off the soon to be deprecated xcode image, that's all we need to do right now.


MrJoy (approved):
LGTM

PR #103, opened 2026-02-21
Include deviceId in ConfigurationData and UserData event payloads

### LINKED ISSUE #92: Include deviceId in ConfigurationData and UserData event payloads
## Summary

A customer needs to target push notifications to specific devices for players with multiple active sessions (multi-device play). To support this, we need to expose our internal `deviceId` through the `OnConfigurationData` and `OnUserData` event payloads so the Unity wrapper can surface it to game code.

## Changes needed

- **`Teak/Events/RemoteConfigurationEvent.m`**: Add `deviceId` to the dictionary returned by `appFacingConfiguration`. The device ID is available via `TeakDeviceConfiguration.deviceId`.
- **`Teak/Events/UserDataEvent.m`**: Add `deviceId` to the dictionary returned by `toDictionary`.

## Acceptance criteria

- `appFacingConfiguration` dictionary includes a `deviceId` key with the device's persistent random ID
- `toDictionary` on `UserDataEvent` includes a `deviceId` key with the same value
- Existing fields in both payloads are unchanged

AlexSc:
## Summary
- Add `deviceId` property to `RemoteConfigurationEvent` and `UserDataEvent`
- `appFacingConfiguration` dictionary now includes `deviceId` (consumed by Unity `OnConfigurationData`)
- `toDictionary` on `UserDataEvent` now includes `deviceId` (consumed by Unity `OnUserData`)
- Device ID sourced from `TeakDeviceConfiguration.deviceId` at each event's dispatch site

Closes #92

## Test plan
- [x] Unit tests for `UserDataEvent.toDictionary` including `deviceId`
- [x] Unit tests for `RemoteConfigurationEvent.appFacingConfiguration` including `deviceId`
- [x] All 18 tests pass (4 new + 14 existing)
- [ ] Integration: verify Unity `ConfigurationData.DeviceId` populates from native payload (blocked on GoCarrot/teak-unity#102)

### DIFF
--- a/Automated/AutomatedTests/RemoteConfigurationEventTests.m
+++ b/Automated/AutomatedTests/RemoteConfigurationEventTests.m
@@ -1,0 +1,47 @@
+#import <XCTest/XCTest.h>
+
+#import "../../Teak/Events/RemoteConfigurationEvent.h"
+#import "../../Teak/Configuration/TeakRemoteConfiguration.h"
+#import "../../Teak/TeakChannelCategory.h"
+
+@import OCHamcrest;
+@import OCMockito;
+
+@interface RemoteConfigurationEventTests : XCTestCase
+@end
+
+@implementation RemoteConfigurationEventTests
+
+- (void)testAppFacingConfigurationIncludesDeviceId {
+  NSString* expectedDeviceId = @"test-device-id-67890";
+
+  TeakRemoteConfiguration* remoteConfig = mock([TeakRemoteConfiguration class]);
+  [given([remoteConfig channelCategories]) willReturn:@[]];
+
+  RemoteConfigurationEvent* event = [[RemoteConfigurationEvent alloc] initWithType:RemoteConfigurationReady];
+  [event setValue:remoteConfig forKey:@"remoteConfiguration"];
+  [event setValue:expectedDeviceId forKey:@"deviceId"];
+
+  NSDictionary* config = [event appFacingConfiguration];
+
+  assertThat(config[@"deviceId"], is(expectedDeviceId));
+  assertThat(config[@"channelCategories"], is(notNilValue()));
+}
+
+- (void)testAppFacingConfigurationPreservesChannelCategories {
+  NSString* deviceId = @"device-xyz";
+
+  TeakRemoteConfiguration* remoteConfig = mock([TeakRemoteConfiguration class]);
+  [given([remoteConfig channelCategories]) willReturn:@[]];
+
+  RemoteConfigurationEvent* event = [[RemoteConfigurationEvent alloc] initWithType:RemoteConfigurationReady];
+  [event setValue:remoteConfig forKey:@"remoteConfiguration"];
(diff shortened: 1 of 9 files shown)


Teak/Events/RemoteConfigurationEvent.m
```diff
@@ -19,7 +21,8 @@ -(nonnull NSDictionary*)appFacingConfiguration {
     [serializedCategories addObject:[category toDictionary]];
   }
   return @{
-    @"channelCategories": serializedCategories
+    @"channelCategories": serializedCategories,
+    @"deviceId": self.deviceId
```
AlexSc: **Inconsistent nil-handling**: This inserts `self.deviceId` directly into the dictionary literal, while `UserDataEvent.m` wraps it with `TeakValueOrNSNull(self.deviceId)`. If `deviceId` is ever nil here (e.g. in tests or a future code path), this will crash with an `NSInvalidArgumentException`.

In practice `TeakDeviceConfiguration` guarantees non-nil, so this is safe today — but the inconsistency between the two events is a footgun. Suggest picking one pattern:
- `TeakValueOrNSNull(self.deviceId)` here too (defensive, matches `UserDataEvent`), or
- Drop `TeakValueOrNSNull` from `UserDataEvent` since both properties are `_Nonnull`

Automated/AutomatedTests/RemoteConfigurationEventTests.m
```diff
@@ -0,0 +1,47 @@
+#import <XCTest/XCTest.h>
+
+#import "../../Teak/Events/RemoteConfigurationEvent.h"
```
AlexSc: Nit: Since `HEADER_SEARCH_PATHS` was added to the test target in this PR, these relative path imports (`../../Teak/...`) could be simplified to just `#import "RemoteConfigurationEvent.h"`. The relative paths are fragile if the test directory structure changes. Same applies to `UserDataEventTests.m`.

Teak/Configuration/TeakRemoteConfiguration.m
```diff
@@ -163,7 +163,7 @@ - (void)configureForSession:(nonnull TeakSession*)session {
                                                     // Categories
                                                     self.channelCategories = [TeakChannelCategory createFromRemoteConfiguration:reply[@"available_categories"]];
 
-                                                    [RemoteConfigurationEvent remoteConfigurationReady:self];
+                                                    [RemoteConfigurationEvent remoteConfigurationReady:self deviceId:[TeakConfiguration configuration].deviceConfiguration.deviceId];
```
AlexSc: I'm inclined to pass the deviceConfiguration through instead of just taking deviceId off of it, to match how we handle remoteConfiguration.

Teak/Events/RemoteConfigurationEvent.m
AlexSc: I would feel better using `TeakValueOrNSNull` here. Even though we're saying it's _Nonnull, we've been bitten many, many times by trying to put nil in a dictionary.

PR #110, opened 2026-02-22
Update Ruby to 3.2.2 and modernize Bundler/CI

### LINKED ISSUE #95: Update Bundler to a non-ancient version.
DependaBot is unhappy with the specified Bundler version.

AlexSc:
## Summary

Closes #95.

- Update Ruby from 2.7.8 (EOL) to 3.2.2, the newest version available on the CI Xcode 14.3.1 image
- Update Bundler from 2.1.4 to 2.4.22
- Remove the sqlite3 `~> 1.6.0` pin (was a Ruby 2.7 workaround); resolves to 1.7.3
- Switch Docker CI jobs (tag_build, deploy_versioned, deploy_latest) from `cimg/ruby:3.1` to `cimg/base:current` — none use Ruby
- Bump CI cache key to avoid stale Ruby 2.7 gem caches
- Replace deprecated `bundle install --path` with `bundle config set --local path`
- Remove `~/.rvm` from CI cache paths (image uses rbenv)

## Test plan

- [x] `bundle exec fastlane test` passes locally (24 tests, 0 failures)
- [ ] CI un-tagged-build workflow passes
- [ ] Verify Docker jobs (tag_build, deploy_versioned, deploy_latest) work with `cimg/base:current`


AlexSc (commented):
## Review

Clean, well-structured PR. The three commits have a logical progression and each is independently revertible. A few observations:

### What looks good

- **Commit ordering is correct.** Ruby/Bundler first, sqlite3 pin removal second, CI config third. Each commit is self-contained and safe to revert independently. Particularly good that the sqlite3 pin removal is separate -- if `1.7.3` causes issues in some environment, it's a single `git revert` away.

- **`cimg/base:current` for Docker jobs.** Confirmed from the issue investigation that `tag_build` uses only git+perl (via the teak/tag-promote orb) and the deploy jobs use only git+awscli. No Ruby dependency, so `cimg/base:current` is the right choice. It's also a smaller image, which should speed up those jobs slightly.

- **Cache key bump.** `v1` -> `v2` avoids any stale Ruby 2.7 gem artifacts contaminating the new environment.

- **`~/.rvm` removal from cache paths.** The CI image uses rbenv, so caching rvm was dead weight.

- **`bundle config set --local path` over `bundle install --path`.** The old flag has been deprecated since Bundler 2.1. The new two-line approach (`config set` then `check || install`) is the modern idiom.

### Minor observations (non-blocking)

1. **`.ruby-version` format**: The file contains `ruby-3.2.2` (with the `ruby-` prefix). Both rbenv and rvm understand this format, so it works. Just noting that the more common rbenv convention is just `3.2.2` without the prefix. Not worth a change since it works either way.

2. **`cimg/base:current` is a rolling tag.** It will track the latest Ubuntu LTS. This is fine for the deploy/tag jobs (they just need git and apt), but worth being aware that the underlying OS could shift from one CI run to the next. If stability is ever a concern, pinning to a specific version (e.g. `cimg/base:2024.01`) is an option. Not a concern for this PR.

3. **`deploy_versioned` uses `attach_workspace` from `build`**, which runs on macOS. The workspace artifacts (`.framework.zip`, `.sha512` files) are platform-independent archives, so attaching them in a Linux Docker job is fine. Just confirming this was already the case before this PR -- no change in behavior.

### CI status

CI is currently running. The test plan appropriately calls out verifying the un-tagged-build workflow and Docker jobs. Wait for CI green before merging.

Overall: well-scoped maintenance PR that addresses issue #95 cleanly. The investigation comments on the issue show good due diligence. Ship it once CI passes.

PR #120, opened 2026-04-14
SDK: push-to-update token forwarding for live activities

AlexSc:
## Summary

Games using iOS Live Activities need to forward per-activity push tokens to the Teak backend so the server can send live activity updates via APNs. Currently the cleanroom prototype observes `Activity.pushTokenUpdates` but has no SDK method to call — it just prints the token.

This adds `+[Teak startedLiveActivity:withToken:]`, which hex-encodes the token and POSTs it to `/me/live_activities` with the activity ID. The game calls this each time a token is received or rotated. The server upserts, so repeated calls update the token rather than creating duplicates.

Blocked by C-605 (the `/me/live_activities` server endpoint), currently in review.

## Key decisions

- Returns `TeakOperation` (not void) so callers can check for `invalid_device` errors — closer to the local notification scheduling pattern than the fire-and-forget push token registration
- No iOS availability guard — the ObjC implementation uses only `NSString`/`NSData`/`TeakRequest`, so version checking is the caller's responsibility (Swift/ActivityKit side)
- No C extern bridge yet — that's C-613 (Unity SDK wrappers), which depends on this

## Test plan

- [x] Unit tests pass (`bundle exec fastlane test` — 32/32, 0 failures)
- [x] Input validation: nil/empty activityId and pushToken return error results
- [x] Request construction: correct endpoint, payload with activity ID and hex-encoded token
- [x] Reply parser: handles `ok` and `invalid_device` responses correctly
- [ ] Integration test with C-605 endpoint once merged
- [ ] Cleanroom app updated to call `Teak.startedLiveActivity()` instead of printing tokens

### DIFF
--- a/Automated/AutomatedTests/LiveActivityTokenForwardingTests.m
+++ b/Automated/AutomatedTests/LiveActivityTokenForwardingTests.m
@@ -1,0 +1,96 @@
+#import <XCTest/XCTest.h>
+
+#import "TeakOperation.h"
+#import <Teak/Teak.h>
+
+@import OCHamcrest;
+@import OCMockito;
+
+@interface LiveActivityTokenForwardingTests : XCTestCase
+@end
+
+@implementation LiveActivityTokenForwardingTests
+
+#pragma mark - Helpers
+
+- (NSData*)sampleTokenData {
+  unsigned char bytes[] = {0xde, 0xad, 0xbe, 0xef, 0xca, 0xfe, 0x01, 0x23};
+  return [NSData dataWithBytes:bytes length:sizeof(bytes)];
+}
+
+- (TeakOperationResult*)runOperationAndGetResult:(TeakOperation*)op {
+  NSOperationQueue* queue = [[NSOperationQueue alloc] init];
+  [queue addOperation:op];
+  [queue waitUntilAllOperationsAreFinished];
+  return (TeakOperationResult*)[op result];
+}
+
+#pragma mark - Input validation: activityId
+
+- (void)testNilActivityIdReturnsErrorOperation {
+  NSData* token = [self sampleTokenData];
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wnonnull"
+  TeakOperation* op = [Teak startedLiveActivity:nil withToken:token];
+#pragma clang diagnostic pop
+
+  XCTAssertNotNil(op, @"Should return an operation even for nil activityId");
+
+  TeakOperationResult* result = [self runOperationAndGetResult:op];
+  XCTAssertNotNil(result);
+  XCTAssertTrue(result.error, @"Result should be an error");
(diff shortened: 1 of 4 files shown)


Teak/Teak.m
```diff
@@ -175,6 +175,57 @@ + (void)setStringProperty:(NSString* _Nonnull)key value:(NSString* _Nonnull)valu
...
+    result = [[TeakOperationResult alloc] initWithStatus:@"error" andErrors:@{@"token" : @[ @"pushToken cannot be null or empty" ]}];
+  }
+
+  if (result) {
+    TeakOperation* op = [TeakOperation withResult:result];
+    [[Teak sharedInstance].operationQueue addOperation:op];
+    return op;
+  }
+
+  if (@available(iOS 16.1, *)) {
```
AlexSc: You know what, I'm looking at it, and I changed my mind. I think I'd prefer not to have this check. I'm a little torn on API_AVAILABLE in the header. I do like the hint that this is an iOS 16.1+ feature and if you aren't targeting that you might be doing something odd, but... we can probably trust to developer and it's not like older versions of iOS are a big target market.

Automated/AutomatedTests/LiveActivityTokenForwardingTests.m
```diff
@@ -0,0 +1,96 @@
...
+  TeakOperationResult* result = [self runOperationAndGetResult:op];
+  XCTAssertNotNil(result);
+  XCTAssertTrue(result.error, @"Result should be an error");
+  XCTAssertEqualObjects(result.status, @"error");
+  XCTAssertNotNil(result.errors[@"token"]);
+}
+
+#pragma mark - Valid inputs
+
+- (void)testValidInputsReturnsNonNilOperation {
```
AlexSc: Can we test that it's making the server request that we expect it to?

AlexSc:
Can you confirm that TeakRequest will include the game_id, api_key, and device_id?

AlexSc:
Yes — `TeakOperation.performRequest:` creates the request via `[TeakRequest requestWithSession:forEndpoint:withPayload:method:callback:]`, which calls through to `initWithSession:...addCommonPayload:YES`. With `addCommonToPayload:YES`, the common payload is merged in at `TeakRequest.m:217-239`:

- `game_id` ← `session.appConfiguration.appId`
- `device_id` ← `session.deviceConfiguration.deviceId`
- `api_key` ← `session.userId`
- `sdk_platform` ← `session.deviceConfiguration.platformString`

Plus `appstore_name`, `sdk_version`, `app_version`, `bundle_id`, `is_sandbox`, etc.

This is the same path used by `TeakNotification` scheduling and channel state operations — all `TeakOperation`-based requests get these params automatically.

AlexSc:
Closing — C-607 was canceled and superseded by C-641 (PR #121, merged 2026-04-17). Code here remains reference material only; no further work planned on this branch.

PR #121, opened 2026-04-17
SDK: startedLiveActivity public API with system_activity_id

AlexSc:
## Summary

- Adds `+[Teak startedLiveActivity:withToken:systemActivityId:]` so games can forward per-activity push-to-update tokens plus the OS-level `system_activity_id` from ActivityKit to the Teak backend, enabling server-driven live activity updates via APNs.
- Hex-encodes the `NSData` token and POSTs `{ live_activity_id, token, system_activity_id }` to `/me/live_activities` via a `TeakOperation`, matching the pattern used by channel state and notification scheduling.
- Supersedes C-607 / PR #120 (closed, discarded). Resolves Linear [C-641](https://linear.app/teakio/issue/C-641/teak-ios-startedliveactivity-public-api-token-system-activity-id).

## Key decisions

- **No `@available(iOS 16.1, *)` guard in ObjC.** Live Activities are iOS 16.1+, but the caller (Swift/ActivityKit) enforces that — duplicating the check in ObjC adds no safety. A header comment documents the requirement.
- **Wire format follows PR #120 + C-637 prose.** The C-637 server handler hasn't merged yet, so fields are `live_activity_id`, `token` (from the existing `/me/live_activities` endpoint), with `system_activity_id` added. Easy to adjust if the server contract diverges.
- **Input validation returns a `TeakOperation` with an error `TeakOperationResult`** rather than nil, keeping the `nonnull` return type and surfacing validation errors via the same path callers use for server errors.
- **Two logical commits** — tests (red), then implementation (green).

## Test plan

- [x] `bundle exec fastlane test` — 36/36 passing (12 new tests)
- [x] Input validation: nil/empty `activityId`, `pushToken`, `systemActivityId` each return an error result with the offending field in `errors`
- [x] Request construction: endpoint is `/me/live_activities`; payload carries `live_activity_id`, lowercase hex-encoded `token`, and `system_activity_id`
- [x] Reply parser: returns non-error for `status: ok`, returns error with `errors` for `status: invalid_device`
- [ ] Integration test against the C-637 endpoint once the server side lands; adjust field names here if the contract diverges
- [ ] Cleanroom app wired up to call `Teak.startedLiveActivity(...)` from its `Activity.pushTokenUpdates` observer (follow-up)

### DIFF
--- a/Teak/Teak.h
+++ b/Teak/Teak.h
@@ -580,9 +580,12 @@ typedef void (^TeakLogListener)(NSString* _Nonnull event,
  * the OS-level system activity identifier to the Teak backend so that server-driven live
  * activity updates can be delivered via APNs.
  *
- * @param activityId       The game-supplied identifier for the live activity instance.
+ * @param activityId       A stable, game-chosen constant string naming the *kind* of live
+ *                         activity (for example, ``@"chest_timer"``). All instances of the
+ *                         same kind of activity should use the same value; Teak uses it as
+ *                         the schedule key for analytics and server-driven updates.
  * @param pushToken        The push-to-update token data from ``Activity.pushTokenUpdates``.
- * @param systemActivityId The OS-level system activity identifier from ``Activity.id``.
+ * @param systemActivityId The OS-level per-instance activity identifier, from ``Activity.id``.
  * @return A TeakOperation which contains the status and result of the call.
  */
 + (nonnull TeakOperation*)startedLiveActivity:(nonnull NSString*)activityId withToken:(nonnull NSData*)pushToken systemActivityId:(nonnull NSString*)systemActivityId;
--- a/Teak/Teak.h
+++ b/Teak/Teak.h
@@ -572,6 +572,25 @@ typedef void (^TeakLogListener)(NSString* _Nonnull event,
  */
 - (nonnull TeakOperation*)setState:(nonnull NSString*)state forChannel:(nonnull NSString*)channel andCategory:(nonnull NSString*)category;
 
+/**
+ * Report a live activity push-to-update token to Teak.
+ *
+ * Call this when a live activity starts and receives its push-to-update token, and again
+ * whenever the token rotates during the activity's lifetime. The SDK forwards the token and
+ * the OS-level system activity identifier to the Teak backend so that server-driven live
+ * activity updates can be delivered via APNs.
+ *
+ * @note Live Activities require iOS 16.1 or later. This is an Objective-C method with no
+ *       ``API_AVAILABLE`` guard; callers in Swift/ActivityKit code are responsible for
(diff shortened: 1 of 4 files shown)


Teak/Teak.h
```diff
@@ -572,6 +572,25 @@ typedef void (^TeakLogListener)(NSString* _Nonnull event,
...
 
+/**
+ * Report a live activity push-to-update token to Teak.
+ *
+ * Call this when a live activity starts and receives its push-to-update token, and again
+ * whenever the token rotates during the activity's lifetime. The SDK forwards the token and
+ * the OS-level system activity identifier to the Teak backend so that server-driven live
+ * activity updates can be delivered via APNs.
+ *
+ * @note Live Activities require iOS 16.1 or later. This is an Objective-C method with no
```
AlexSc: Nothing fails if we’re on an older os, yea?

Teak/Teak.h
```diff
@@ -572,6 +572,25 @@ typedef void (^TeakLogListener)(NSString* _Nonnull event,
...
+ *       ``API_AVAILABLE`` guard; callers in Swift/ActivityKit code are responsible for
+ *       enforcing the availability check before invoking it.
+ *
+ * @param activityId       The game-supplied identifier for the live activity instance.
```
AlexSc: Let’s make it clear that this should be a constant value, e.g. “chest_timer”

Teak/Teak.m
```diff
@@ -175,6 +175,56 @@ + (void)setStringProperty:(NSString* _Nonnull)key value:(NSString* _Nonnull)valu
   [[self sharedInstance] setStringAttribute:value forKey:key];
 }
 
+// Live Activities require iOS 16.1+; the caller (Swift/ActivityKit) is responsible for
+// gating invocation on platform availability rather than duplicating the check here.
```
AlexSc: Since we don’t fail comment is unnecessary

Teak/Teak.m
```diff
@@ -175,6 +175,56 @@ + (void)setStringProperty:(NSString* _Nonnull)key value:(NSString* _Nonnull)valu
...
++ (nonnull TeakOperation*)startedLiveActivity:(nonnull NSString*)activityId withToken:(nonnull NSData*)pushToken systemActivityId:(nonnull NSString*)systemActivityId {
```
AlexSc, line 178: Will this wait for user id to be ready? If it matches local notifications we’re probably fine, but double checking

AlexSc: Yes — `TeakOperation.performRequest:` routes through `[TeakSession whenUserIdIsOrWasReadyRun:...]` (TeakOperation.m:117) before dispatching the `TeakRequest`, same as local notification scheduling. If this is called before the session has identified the user, it waits.

PR #122, opened 2026-04-17
SDK: scheduleLiveActivityUpdate + cancelLiveActivityUpdates public APIs

AlexSc:
## Summary
- Adds `+scheduleLiveActivityUpdate:sendTime:customData:systemData:` on `Teak` — game calls this between `+startedLiveActivity:...` and activity end to schedule a single update at a future `send_time`. `customData` becomes APNs `content-state`; optional `systemData` carries Apple system fields.
- Adds `+cancelLiveActivityUpdates:` on `Teak` — cancels all pending updates for a given `live_activity_id` and surfaces the server's canceled count via a new `TeakOperationLiveActivityCancelResult`.
- Both target the `/me/live_activity_updates` and `/me/cancel_all_live_activity_updates` endpoints that landed in [Carrot C-609](https://linear.app/teakio/issue/C-609).

Closes [C-643](https://linear.app/teakio/issue/C-643/teak-ios-scheduleliveactivityupdate-cancelliveactivityupdates-public).

## Key decisions
- **No `@available(iOS 16.1, *)` guard in ObjC.** Matches the [C-641 / PR #121](https://github.com/GoCarrot/teak-ios/pull/121) precedent that just merged: the Swift/ActivityKit caller is responsible for gating on availability, and the header documents the iOS 16.1 requirement. The issue's AC calls for in-SDK guards + no-op results below 16.1; explicitly deviating to match what shipped for the sibling API.
- **Cancel surfaces the canceled count.** New `TeakOperationLiveActivityCancelResult` subclass (`NSInteger canceled`) so callers can read it without parsing the raw reply dict. Schedule returns a plain `TeakOperationResult` — the update id isn't exposed, since the cancel API operates on `live_activity_id` and no caller use case for the per-update id exists today.
- **Endpoint paths from the landed server code, not the issue description.** The issue text says `/me/live_activity_updates/cancel_all`; what actually landed in C-609 is `/me/cancel_all_live_activity_updates`. SDK targets the landed path. Noted on the Linear issue too.
- **`custom_data` / `system_data` sent as JSON strings.** The server stores them as text (feature spec in `carrot` sends them as JSON strings). SDK serializes the input dicts with `NSJSONSerialization`; invalid input (non-JSON types like `NSDate`) returns an immediate error result without touching the network.
- **`nil` systemData is allowed.** Server treats `system_data` as optional. SDK sends `NSNull` when nil, matching the `TeakNotification` `personalization_data` precedent.

## Test plan
- [x] `bundle exec fastlane test` — full suite green (16 new schedule tests, 7 new cancel tests, all existing suites still 0 failures).
- Schedule coverage: nil/empty activityId; nil sendTime; nil customData; non-JSON-serializable customData; non-JSON-serializable systemData; happy-path request construction (endpoint, `live_activity_id`, `send_time` unix-epoch integer, `custom_data` round-trips through JSON, `system_data` round-trips through JSON); nil systemData → absent/`NSNull`; reply parser for `ok` / `error` / `invalid_device`.
- Cancel coverage: nil/empty activityId; happy-path request construction (endpoint, `live_activity_id`); reply parser for `ok` with non-zero and zero canceled counts, and `invalid_device`.
- [ ] Integration-test against a live Carrot build once the game-side Swift/ActivityKit integration lands (out of scope for this PR).

### DIFF
--- a/Automated/AutomatedTests/LiveActivityUpdateCancellationTests.m
+++ b/Automated/AutomatedTests/LiveActivityUpdateCancellationTests.m
@@ -1,0 +1,121 @@
+#import <XCTest/XCTest.h>
+
+#import "TeakOperation.h"
+#import <Teak/Teak.h>
+
+@import OCHamcrest;
+@import OCMockito;
+
+// Re-expose internal replyParser property for testing
+@interface TeakOperation ()
+@property (nonatomic, copy, nullable) id (^replyParser)(NSDictionary* _Nonnull);
+@end
+
+@interface LiveActivityUpdateCancellationTests : XCTestCase
+@end
+
+@implementation LiveActivityUpdateCancellationTests
+
+#pragma mark - Helpers
+
+- (TeakOperationResult*)runOperationAndGetResult:(TeakOperation*)op {
+  NSOperationQueue* queue = [[NSOperationQueue alloc] init];
+  [queue addOperation:op];
+  [queue waitUntilAllOperationsAreFinished];
+  return (TeakOperationResult*)[op result];
+}
+
+- (NSDictionary*)requestParamsFromOperation:(TeakOperation*)op {
+  NSInvocation* inv = op.invocation;
+  __unsafe_unretained NSDictionary* requestParams;
+  [inv getArgument:&requestParams atIndex:2];
+  return requestParams;
+}
+
+- (TeakOperation*)validOperation {
+  return [Teak cancelLiveActivityUpdates:@"chest_timer"];
+}
+
+#pragma mark - Input validation
+
+- (void)testNilActivityIdReturnsErrorOperation {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wnonnull"
+  TeakOperation* op = [Teak cancelLiveActivityUpdates:nil];
+#pragma clang diagnostic pop
+
+  XCTAssertNotNil(op);
+
+  TeakOperationResult* result = [self runOperationAndGetResult:op];
+  XCTAssertNotNil(result);
+  XCTAssertTrue(result.error);
+  XCTAssertEqualObjects(result.status, @"error");
+  XCTAssertNotNil(result.errors[@"activityId"]);
+}
+
+- (void)testEmptyActivityIdReturnsErrorOperation {
+  TeakOperation* op = [Teak cancelLiveActivityUpdates:@""];
+
+  XCTAssertNotNil(op);
+
+  TeakOperationResult* result = [self runOperationAndGetResult:op];
+  XCTAssertNotNil(result);
+  XCTAssertTrue(result.error);
+  XCTAssertEqualObjects(result.status, @"error");
+  XCTAssertNotNil(result.errors[@"activityId"]);
+}
+
+#pragma mark - Request construction
+
+- (void)testRequestUsesCorrectEndpoint {
+  TeakOperation* op = [self validOperation];
+
+  NSDictionary* requestParams = [self requestParamsFromOperation:op];
+  XCTAssertEqualObjects(requestParams[@"endpoint"], @"/me/cancel_all_live_activity_updates");
+}
+
+- (void)testRequestPayloadContainsActivityId {
+  TeakOperation* op = [self validOperation];
+
+  NSDictionary* payload = [self requestParamsFromOperation:op][@"payload"];
+  XCTAssertEqualObjects(payload[@"live_activity_id"], @"chest_timer");
+}
+
+#pragma mark - Reply parser
+
+- (void)testReplyParserSurfacesCanceledCountForOk {
+  TeakOperation* op = [self validOperation];
+
+  NSDictionary* reply = @{@"status" : @"ok", @"canceled" : @3};
+  TeakOperationLiveActivityCancelResult* result =
+      (TeakOperationLiveActivityCancelResult*)op.replyParser(reply);
+
+  XCTAssertFalse(result.error);
+  XCTAssertEqualObjects(result.status, @"ok");
+  XCTAssertTrue([result isKindOfClass:[TeakOperationLiveActivityCancelResult class]],
+                @"Reply parser should return a TeakOperationLiveActivityCancelResult");
+  XCTAssertEqual(result.canceled, 3);
+}
+
+- (void)testReplyParserZeroCanceledIsOk {
+  TeakOperation* op = [self validOperation];
+
+  NSDictionary* reply = @{@"status" : @"ok", @"canceled" : @0};
+  TeakOperationLiveActivityCancelResult* result =
+      (TeakOperationLiveActivityCancelResult*)op.replyParser(reply);
+
+  XCTAssertFalse(result.error);
+  XCTAssertEqual(result.canceled, 0);
+}
+
+- (void)testReplyParserReturnsErrorForInvalidDevice {
+  TeakOperation* op = [self validOperation];
+
+  NSDictionary* reply = @{@"status" : @"invalid_device", @"device_id" : @"nonexistent"};
+  TeakOperationResult* result = op.replyParser(reply);
+
+  XCTAssertTrue(result.error);
+  XCTAssertEqualObjects(result.status, @"invalid_device");
+}
+
+@end
--- a/Automated/AutomatedTests/LiveActivityUpdateCancellationTests.m
+++ b/Automated/AutomatedTests/LiveActivityUpdateCancellationTests.m
@@ -1,36 +1,15 @@
-#import <XCTest/XCTest.h>
-
-#import "TeakOperation.h"
+#import "TeakOperationTestCase.h"
 #import <Teak/Teak.h>
 
 @import OCHamcrest;
 @import OCMockito;
 
-// Re-expose internal replyParser property for testing
-@interface TeakOperation ()
-@property (nonatomic, copy, nullable) id (^replyParser)(NSDictionary* _Nonnull);
-@end
-
-@interface LiveActivityUpdateCancellationTests : XCTestCase
+@interface LiveActivityUpdateCancellationTests : TeakOperationTestCase
 @end
 
 @implementation LiveActivityUpdateCancellationTests
 
 #pragma mark - Helpers
-
-- (TeakOperationResult*)runOperationAndGetResult:(TeakOperation*)op {
-  NSOperationQueue* queue = [[NSOperationQueue alloc] init];
-  [queue addOperation:op];
-  [queue waitUntilAllOperationsAreFinished];
-  return (TeakOperationResult*)[op result];
-}
-
-- (NSDictionary*)requestParamsFromOperation:(TeakOperation*)op {
-  NSInvocation* inv = op.invocation;
-  __unsafe_unretained NSDictionary* requestParams;
-  [inv getArgument:&requestParams atIndex:2];
-  return requestParams;
-}
 
 - (TeakOperation*)validOperation {
   return [Teak cancelLiveActivityUpdates:@"chest_timer"];
--- a/Automated/AutomatedTests/LiveActivityUpdateCancellationTests.m
+++ b/Automated/AutomatedTests/LiveActivityUpdateCancellationTests.m
@@ -1,13 +1,21 @@
-#import "TeakOperationTestCase.h"
+#import <XCTest/XCTest.h>
+
+#import "TeakOperationTestDriver.h"
 #import <Teak/Teak.h>
 
 @import OCHamcrest;
 @import OCMockito;
 
-@interface LiveActivityUpdateCancellationTests : TeakOperationTestCase
+@interface LiveActivityUpdateCancellationTests : XCTestCase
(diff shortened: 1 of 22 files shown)


Teak/Teak.h
```diff
@@ -590,6 +590,54 @@ typedef void (^TeakLogListener)(NSString* _Nonnull event,
...
+ * Schedule a single update to be delivered to a live activity at a future time.
+ *
+ * The game calls this between ``+startedLiveActivity:withToken:systemActivityId:`` and the
+ * activity ending. The scheduled update's ``customData`` is delivered as APNs
+ * ``content-state``; ``systemData`` carries Apple system fields (``event``, ``stale-date``,
+ * ``dismissal-date``). Server-side, ``send_time`` is rejected if it falls outside a 10-hour
+ * window from the live activity's creation; idempotency applies as it does for other
+ * Teak-scheduled sends.
+ *
+ * @note Live Activities require iOS 16.1 or later. This is an Objective-C method with no
```
AlexSc: Unnecessary

Teak/Teak.h
```diff
@@ -590,6 +590,54 @@ typedef void (^TeakLogListener)(NSString* _Nonnull event,
...
+                                          customData:(nonnull NSDictionary*)customData
+                                          systemData:(nullable NSDictionary*)systemData;
+
+/**
+ * Cancel all pending scheduled updates for a live activity.
+ *
+ * Scopes to the current user's updates for ``activityId``; updates scheduled for other
+ * users or other activity kinds are unaffected.
+ *
+ * @note Live Activities require iOS 16.1 or later. This is an Objective-C method with no
```
AlexSc: Unnecessary

Teak/Teak.h
```diff
@@ -590,6 +590,54 @@ typedef void (^TeakLogListener)(NSString* _Nonnull event,
  */
 + (nonnull TeakOperation*)startedLiveActivity:(nonnull NSString*)activityId withToken:(nonnull NSData*)pushToken systemActivityId:(nonnull NSString*)systemActivityId;
 
+/**
+ * Schedule a single update to be delivered to a live activity at a future time.
+ *
+ * The game calls this between ``+startedLiveActivity:withToken:systemActivityId:`` and the
+ * activity ending. The scheduled update's ``customData`` is delivered as APNs
+ * ``content-state``; ``systemData`` carries Apple system fields (``event``, ``stale-date``,
+ * ``dismissal-date``). Server-side, ``send_time`` is rejected if it falls outside a 10-hour
```
AlexSc: Weird segue, cut.

Teak/Teak.h
```diff
@@ -590,6 +590,54 @@ typedef void (^TeakLogListener)(NSString* _Nonnull event,
...
+ * Teak-scheduled sends.
+ *
+ * @note Live Activities require iOS 16.1 or later. This is an Objective-C method with no
+ *       ``API_AVAILABLE`` guard; callers in Swift/ActivityKit code are responsible for
+ *       enforcing the availability check before invoking it.
+ *
+ * @param activityId A stable, game-chosen constant string naming the *kind* of live
+ *                   activity (for example, ``@"chest_timer"``) — the same value that was
+ *                   passed to ``+startedLiveActivity:withToken:systemActivityId:``.
+ * @param sendTime   Absolute delivery time for the update.
```
AlexSc: Send time should be within 8 hours of activity start

Teak/Teak.m
```diff
@@ -223,6 +223,127 @@ + (nonnull TeakOperation*)startedLiveActivity:(nonnull NSString*)activityId with
...
+  TeakLog_t(@"[Teak scheduleLiveActivityUpdate]", @{@"activityId" : _(activityId), @"offset" : @(offset)});
+
+  TeakOperationResult* result = nil;
+
+  if (activityId == nil || activityId.length == 0) {
+    TeakLog_e(@"live_activity.schedule.error", @"activityId cannot be null or empty");
+    result = [[TeakOperationResult alloc] initWithStatus:@"error" andErrors:@{@"activityId" : @[ @"activityId cannot be null or empty" ]}];
+  }
+
+  if (!result && offset <= 0) {
```
AlexSc: We’re allowing 0 offset

Teak/Teak.m
```diff
@@ -223,6 +223,127 @@ + (nonnull TeakOperation*)startedLiveActivity:(nonnull NSString*)activityId with
...
+    *outResult = [[TeakOperationResult alloc] initWithStatus:@"error" andErrors:@{fieldName : @[ errorMessage ]}];
+    return nil;
+  }
+  return [[NSString alloc] initWithData:jsonData encoding:NSUTF8StringEncoding];
+}
+
+// Live Activities require iOS 16.1+; the caller (Swift/ActivityKit) is responsible for
+// gating invocation on platform availability rather than duplicating the check here.
++ (nonnull TeakOperation*)scheduleLiveActivityUpdate:(nonnull NSString*)activityId
+                                              offset:(NSTimeInterval)offset
```
AlexSc: Does this type match local notifications? Internally we’re truncating— especially if local notifs differ the type should be changed to match

Automated/AutomatedTests/LiveActivityUpdateCancellationTests.m
```diff
@@ -0,0 +1,121 @@
...
+@end
+
+@interface LiveActivityUpdateCancellationTests : XCTestCase
+@end
+
+@implementation LiveActivityUpdateCancellationTests
+
+#pragma mark - Helpers
+
+- (TeakOperationResult*)runOperationAndGetResult:(TeakOperation*)op {
```
AlexSc: This feels like it should be a shared helper

Automated/AutomatedTests/LiveActivityUpdateCancellationTests.m
```diff
@@ -0,0 +1,100 @@
+#import "TeakOperationTestCase.h"
+#import <Teak/Teak.h>
+
+@import OCHamcrest;
+@import OCMockito;
+
+@interface LiveActivityUpdateCancellationTests : TeakOperationTestCase
```
AlexSc: Do we have an option for composition instead of inheritance? What would that look like, and would it be any cleaner?

AlexSc: Three non-inheritance shapes, in rough order of how close they are to the current PR:

**1. Free functions in a shared header** — the lightest touch. `TeakOperationTestHelpers.h` keeps the `@interface TeakOperation ()` re-decl plus two C-ish helpers:

```objc
TeakOperationResult* TeakOperationTestRunSync(TeakOperation* op);
NSDictionary* TeakOperationTestRequestParams(TeakOperation* op);
```

Test classes stay `XCTestCase`; each callsite becomes `TeakOperationTestRunSync(op)` instead of `[self runOperationAndGetResult:op]`.

**2. Category on `XCTestCase`** — test classes still inherit from `XCTestCase`, but pick up the helpers from a category in the shared header:

```objc
@interface XCTestCase (TeakOperationHelpers)
- (TeakOperationResult*)teak_runOperationAndGetResult:(TeakOperation*)op;
- (NSDictionary*)teak_requestParamsFromOperation:(TeakOperation*)op;
@end
```

Reads closer to the current code but swaps `XCTestCase` for `TeakOperationTestCase` without forcing the subclass.

**3. Category on `TeakOperation`** — `[op teak_test_runSync]` / `[op teak_test_requestParams]`. Most object-oriented but pollutes `TeakOperation`'s method namespace even at test time.

My read: cleaner is a stretch. The current base-class version is two helpers and a private-property re-decl that all exist specifically to unit-test `TeakOperation` — packaging them as a subclass says that out loud. Composition variants pay a naming-prefix tax (`teak_`, `TeakOperationTest…`) to avoid one inheritance arrow, and none of them get shorter at the callsite. The one real limitation of inheritance is that a test class can't simultaneously inherit from something else (e.g., a future `TeakNetworkingTestCase`) — today there's no such class, but if/when one shows up, shape (2) is the easy migration: drop the `: TeakOperationTestCase`, keep the method calls, move the bodies to a category.

Happy to flip to (1) or (2) if you'd rather — just want to flag that I don't think either is a win on its own terms, only as insurance against a future multiple-inheritance squeeze.

AlexSc: What would standard best practices for this be? I do want to plan ahead to make it easier to compose our unit test helpers.

AlexSc: Standard iOS best-practice answer, as of the Swift Testing era, has actually moved *away* from XCTestCase inheritance for exactly the forward-composition reason you're raising. Apple's Swift Testing framework (2024) has no base class at all — just `@Test` functions with traits — which is the strongest signal that inheritance for test infrastructure is end-of-life thinking even inside Apple. For ObjC/XCTest specifically, the compositional patterns that scale are, in order:

**1. Helper *objects* on a test `@property`** — "the test *has a* helper." For our case:

```objc
// TeakOperationTestDriver.h
@interface TeakOperation ()
@property (nonatomic, copy, nullable) id (^replyParser)(NSDictionary* _Nonnull);
@end

@interface TeakOperationTestDriver : NSObject
- (TeakOperationResult*)runSync:(TeakOperation*)op;
- (NSDictionary*)requestParamsFor:(TeakOperation*)op;
@end
```

Each test:
```objc
@interface LiveActivityUpdateCancellationTests : XCTestCase
@property (nonatomic) TeakOperationTestDriver* driver;
@end

- (void)setUp {
  [super setUp];
  self.driver = [[TeakOperationTestDriver alloc] init];
}

- (void)testFoo {
  TeakOperation* op = ...;
  NSDictionary* payload = [self.driver requestParamsFor:op][@"payload"];
}
```

Cost: one line in `setUp`, a small `self.driver.` prefix at callsites. Benefit that *actually* composes: when `TeakSessionMock`, `TeakLogCapture`, or `TeakRequestInterceptor` show up later, they stack as additional `@property`s on the same test with independent APIs. Inheritance piles everything onto one base class and can't stack; categories on `XCTestCase` stack flatly but share the `self` namespace (so `teak_foo` / `other_foo` conventions leak into every test's method list). Helper objects keep each bundle of state-and-behavior in its own named spot, which is exactly the property you want when planning for "more helpers later."

**2. Categories on `XCTestCase`** — fine for stateless, stateless-looking utility like "run this op synchronously." Limits: no `@property` hooks for teardown/state, and the prefix-collision risk as the set grows.

**3. Free functions** — fine for pure transforms (`requestParamsFor(op)` genuinely is one). Doesn't compose beyond that.

If the goal is "make composing helpers easier later," I'd flip this PR's base class to shape (1): one `TeakOperationTestDriver` helper, held by each test as a property. The existing helper extraction then becomes the template for the next helper we write, instead of the one-and-only base class that future helpers have to either elbow into or work around.

Want me to make that flip on this PR, or keep it scoped and thread the reshape through a follow-up once we have a second helper driving the concrete shape?

AlexSc: Let's do the flip. I'd say "file a followup issue", but we're going to lose it and end up with three competing standards before we sit down and do it right. Let's land the template the way we want it used so that we at least come back to three mostly similar reimplementations to clean up instead of three different ones.

AlexSc: Done — d64fd6e. `TeakOperationTestCase` → `TeakOperationTestDriver`; each test now holds `self.driver` and calls `[self.driver runSync:op]` / `[self.driver requestParamsFor:op]`. All three live-activity suites are back to inheriting `XCTestCase`. Full test suite still green (12/18/7).

AlexSc:
Quick fast follow check -- are there any other test cases that could use our new driver?

And can you resolve the 4.3-stable merge conflicts?

AlexSc:
Both items handled:

**Other driver candidates — none today.** Audited the suite (`grep -l TeakOperation\|NSOperationQueue\|replyParser`) plus the two new 4.3-stable tests:
- `PushToStartTokenRegistrationTests` — goes through `TeakEvent` / `PushRegistrationEvent`, the public API is `+registerPushToStartToken:` returning `void`. No `TeakOperation` touched, no fit for the driver.
- `TeakDeviceConfigurationTests` — pure object-state unit tests, no network / operation plumbing.

There *are* three other `TeakOperation`-returning public APIs that could pick up the driver if/when they get tests: `-setState:forChannel:`, `-setState:forChannel:andCategory:`, `+scheduleNotificationForCreative:secondsFromNow:personalizationData:`. None have tests today, so there's nothing to migrate — just noting them as natural first users next time we expand coverage there.

**Merge conflicts resolved in 3d36b8d.** `Teak.m` auto-merged cleanly. `Teak.h` and `Automated/Automated.xcodeproj/project.pbxproj` were union-resolves — both sides appended in the same section, kept both. No semantic drift. Full suite post-merge: 12 token-forwarding + 18 schedule + 7 cancel + 4 push-to-start-registration + 8 device-configuration, 0 failures.

PR #123, opened 2026-04-18
C-608: Add registerPushToStartToken: public API

AlexSc:
## Summary
- Adds `+[Teak registerPushToStartToken:]` so games can forward the live activity push-to-start token from `Activity<T>.pushToStartTokenUpdates` to Teak.
- Stores the token on `TeakDeviceConfiguration` and sends it to the server as `live_activity_push_to_start_key` in the `users.json` identify payload, mirroring the existing `apns_push_key` flow.
- Server side (C-606) already accepts this field and records it as a new `push_type`.

## Key decisions
- Reused `PushRegistrationEvent` with a new `LiveActivityPushToStartRegistered` event type rather than creating a new event class, matching how `LifecycleEvent` handles multiple lifecycle transitions. Avoids churning the Teak.xcodeproj across its two framework targets.
- Stored in memory (not `NSUserDefaults`), just like `pushToken` — the game calls this at each launch after observing `pushToStartTokenUpdates`.
- KVO on `liveActivityPushToStartToken` triggers `identifyUserInfoHasChanged` so token rotation during a session re-sends the identify request.
- `live_activity_push_to_start_key` is only included in the payload when non-empty (additive), whereas `apns_push_key` always writes `""` when unset. This avoids ambiguously signaling \"clear\" on launches before the game has registered a token.
- No iOS version guard in the public API — safe to call on any iOS version, as the game just won't invoke it on OS versions that don't produce a token.

## Test plan
- [x] `bundle exec fastlane test` (48 tests passing, including 12 new)
- New unit coverage:
  - `PushToStartTokenRegistrationTests`: public API nil/empty/valid inputs and event posting (type + lowercase hex token)
  - `TeakDeviceConfigurationTests`: initial empty state, event routing between `pushToken` and `liveActivityPushToStartToken`, token rotation, `to_h` payload serialization under `pushRegistration`
- [ ] Manual: wire into teak-swift-cleanroom, observe `live_activity_push_to_start_key` in the next `users.json` request after calling `Teak.registerPushToStartToken(_:)` with an `Activity.pushToStartTokenUpdates` value

Closes C-608.

### DIFF
--- a/Teak/Configuration/TeakDeviceConfiguration.m
+++ b/Teak/Configuration/TeakDeviceConfiguration.m
@@ -189,6 +189,9 @@ NSString* const TeakDeviceConfiguration_NotificationDisplayState_NotDetermined =
       if (self.pushToken != nil) self.pushToken = nil;
     } break;
     case LiveActivityPushToStartRegistered: {
+      // No symmetric "unregister" case: push-to-start tokens are per-device and persistent.
+      // The OS rotates the token in place via Activity.pushToStartTokenUpdates; it doesn't
+      // revoke one the way push permission revocation nils out pushToken.
       NSString* token = ((PushRegistrationEvent*)event).token;
       // Check before assignment so KVO isn't triggered unless it should be
       if (self.liveActivityPushToStartToken == nil || ![self.liveActivityPushToStartToken isEqualToString:token]) {
--- a/Teak/Configuration/TeakDeviceConfiguration.m
+++ b/Teak/Configuration/TeakDeviceConfiguration.m
@@ -18,6 +18,7 @@ NSString* const TeakDeviceConfiguration_NotificationDisplayState_NotDetermined =
 @property (strong, nonatomic, readwrite) NSString* deviceId;
 @property (strong, nonatomic, readwrite) NSString* deviceModel;
 @property (strong, nonatomic, readwrite) NSString* pushToken;
+@property (strong, nonatomic, readwrite) NSString* liveActivityPushToStartToken;
 @property (strong, nonatomic, readwrite) NSString* platformString;
 @property (strong, nonatomic, readwrite) NSString* advertisingIdentifier;
 @property (strong, nonatomic, readwrite) NSString* notificationDisplayEnabled;
@@ -89,6 +90,7 @@ NSString* const TeakDeviceConfiguration_NotificationDisplayState_NotDetermined =
     // Make sure these are not nil
     self.advertisingIdentifier = @"";
     self.pushToken = @"";
+    self.liveActivityPushToStartToken = @"";
 
     [TeakEvent addEventHandler:self];
 
@@ -152,7 +154,8 @@ NSString* const TeakDeviceConfiguration_NotificationDisplayState_NotDetermined =
     @"limitAdTracking" : [NSNumber numberWithBool:self.limitAdTracking],
     @"notificationDisplayEnabled" : self.notificationDisplayEnabled,
     @"pushRegistration" : @{
-      @"apns_push_key" : TeakValueOrNSNull(self.pushToken)
+      @"apns_push_key" : TeakValueOrNSNull(self.pushToken),
+      @"live_activity_push_to_start_key" : TeakValueOrNSNull(self.liveActivityPushToStartToken)
     }
   };
 }
@@ -184,6 +187,13 @@ NSString* const TeakDeviceConfiguration_NotificationDisplayState_NotDetermined =
     } break;
     case PushUnRegistered: {
       if (self.pushToken != nil) self.pushToken = nil;
+    } break;
+    case LiveActivityPushToStartRegistered: {
+      NSString* token = ((PushRegistrationEvent*)event).token;
+      // Check before assignment so KVO isn't triggered unless it should be
+      if (self.liveActivityPushToStartToken == nil || ![self.liveActivityPushToStartToken isEqualToString:token]) {
+        self.liveActivityPushToStartToken = token;
+      }
     } break;
     case LifecycleActivate: {
       [self updateValuesThatCouldHaveChanged];
--- a/Teak/Configuration/TeakDeviceConfiguration.m
+++ b/Teak/Configuration/TeakDeviceConfiguration.m
@@ -189,9 +189,6 @@ NSString* const TeakDeviceConfiguration_NotificationDisplayState_NotDetermined =
       if (self.pushToken != nil) self.pushToken = nil;
     } break;
     case LiveActivityPushToStartRegistered: {
-      // No symmetric "unregister" case: push-to-start tokens are per-device and persistent.
-      // The OS rotates the token in place via Activity.pushToStartTokenUpdates; it doesn't
-      // revoke one the way push permission revocation nils out pushToken.
       NSString* token = ((PushRegistrationEvent*)event).token;
       // Check before assignment so KVO isn't triggered unless it should be
       if (self.liveActivityPushToStartToken == nil || ![self.liveActivityPushToStartToken isEqualToString:token]) {
--- a/Teak/Teak.h
+++ b/Teak/Teak.h
@@ -597,9 +597,6 @@ typedef void (^TeakLogListener)(NSString* _Nonnull event,
  * ``Activity<Attributes>.pushToStartTokenUpdates`` — typically at launch and again
  * whenever the OS emits a rotated token.
  *
- * Safe to call on any iOS version. Push-to-start tokens are an iOS 17.2+ feature,
- * so in practice this is only invoked on iOS 17.2 and newer.
- *
  * @param token The push-to-start token data from ``Activity.pushToStartTokenUpdates``.
  */
 + (void)registerPushToStartToken:(nonnull NSData*)token;
--- a/Teak/Teak.h
+++ b/Teak/Teak.h
@@ -598,8 +598,8 @@ typedef void (^TeakLogListener)(NSString* _Nonnull event,
  * activities for this app; Teak stores it on the device configuration and forwards it with the
  * next session identify request.
  *
- * Safe to call on any iOS version: on versions below iOS 17.2 the OS never produces a
- * push-to-start token, so this method simply won't be invoked.
+ * Safe to call on any iOS version. In practice callers will only invoke this on iOS 17.2+,
+ * since the OS does not produce a push-to-start token on older versions.
  *
  * @param token The push-to-start token data from ``Activity.pushToStartTokenUpdates``.
  */
--- a/Teak/Teak.h
+++ b/Teak/Teak.h
@@ -590,6 +590,21 @@ typedef void (^TeakLogListener)(NSString* _Nonnull event,
  */
 + (nonnull TeakOperation*)startedLiveActivity:(nonnull NSString*)activityId withToken:(nonnull NSData*)pushToken systemActivityId:(nonnull NSString*)systemActivityId;
 
+/**
+ * Report the live activity push-to-start token for this device.
+ *
+ * Call this once at launch, after observing ``Activity<Attributes>.pushToStartTokenUpdates``, and
+ * whenever the OS rotates the token. The token is per-device and applies to all push-to-start
+ * activities for this app; Teak stores it on the device configuration and forwards it with the
+ * next session identify request.
+ *
+ * Safe to call on any iOS version: on versions below iOS 17.2 the OS never produces a
+ * push-to-start token, so this method simply won't be invoked.
+ *
+ * @param token The push-to-start token data from ``Activity.pushToStartTokenUpdates``.
+ */
++ (void)registerPushToStartToken:(nonnull NSData*)token;
+
 @end
 
 #endif /* __OBJC__ */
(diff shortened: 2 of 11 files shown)


AlexSc:
## Code Review: PR #123 — `registerPushToStartToken:`

### Overview
Adds a new Teak public API `+[Teak registerPushToStartToken:]` that lets games forward the per-device live activity push-to-start token (from `Activity<T>.pushToStartTokenUpdates`) to Teak. Token is stored on `TeakDeviceConfiguration.liveActivityPushToStartToken`, observed via KVO by `TeakSession` to trigger a re-identify, and serialized as `live_activity_push_to_start_key` in the `users.json` payload. Reuses `PushRegistrationEvent` with a new `LiveActivityPushToStartRegistered` event type, mirroring how `apns_push_key` / `pushToken` already work. 12 new unit tests.

### Strengths
- Design is tight and consistent: follows the existing `PushRegistered` → `TeakDeviceConfiguration.pushToken` → KVO-on-session → `identifyUserInfoHasChanged` pattern exactly. Reusing `PushRegistrationEvent` (rather than a new event class) is the right call and matches the `LifecycleEvent` precedent noted in the description.
- Public API: docstring is clear, correctly notes that it's safe on older iOS, and uses `_Nonnull` consistently.
- KVO guard in `TeakDeviceConfiguration.handleEvent:` (`isEqualToString:` check before assignment) mirrors `PushRegistered` and avoids spurious identify re-sends on duplicate events.
- Payload conditional in `TeakSession.sendUserIdentifier` respects `dataCollectionConfiguration.enablePushKey` and only writes the key when non-empty — avoids ambiguous "clear" semantics before the game has observed a token. Justification in the PR description is sound.
- Hex conversion uses the existing `TeakHexStringFromData` helper, so lowercase-hex and empty-input handling stays consistent with the APNs flow.
- Tests cover the right surface: nil/empty rejection, correct event type, lowercase hex encoding, routing between `PushRegistered` and `LiveActivityPushToStartRegistered` (both directions), token rotation, idempotent same-token, and `to_h` serialization — both alone and alongside `apns_push_key`.

### Suggestions / Points to consider

**1. Log message for nil/empty token is imprecise.** In `Teak.m:181`:
```
TeakLog_e(@"live_activity.push_to_start.error", @"token cannot be null or empty");
```
The caller violated `_Nonnull` to hit this, but more importantly the `nil` case is the only time `TeakHexStringFromData` returns `nil` aside from zero-length data. Compare to `+startedLiveActivity:` (line 195-199) which separates the two cases and attaches the error to a structured field. Not blocking, but a slightly more specific message (or distinguishing `nil` vs. 0-length) would read better in logs.

**2. No `unregister` / token-clearing path.** `PushRegistered` has a matching `PushUnRegistered` that nils out `pushToken` when the OS revokes push permission. There's no equivalent for push-to-start. That's probably correct today (live activity tokens don't get explicitly "revoked" the way APNs can be), and the PR description calls it out as intentional. Worth a one-line comment in `TeakDeviceConfiguration.m` next to the `LiveActivityPushToStartRegistered` case explaining why there's no symmetric clear case, so a future reader doesn't wonder if it's an oversight.

**3. `dispatchUserDataEvent` is unchanged.** `TeakSession.m:896-913` builds a `pushRegistration` dictionary for `UserDataEvent` using only `apns`. If downstream consumers of `TeakUserData.pushRegistration` would also want `live_activity_push_to_start_key`, this is a gap. If intentionally scoped out (server identify is all that's needed for C-606), worth a note in the PR or a follow-up ticket.

**4. Test — private-API coupling.** `PushToStartTokenRegistrationTests.m` declares a `TeakEvent (Testing)` category exposing `+eventProcessingQueue` and uses `dispatch_sync` on it as a barrier. This is clever but couples the test to an internal implementation detail (the queue's existence, its serial nature, and the method's presence). Two lighter alternatives:
- Use `XCTestExpectation` + `fulfill` inside the block handler with a short timeout — robust to the queue being replaced or renamed.
- Or assert on the `TeakDeviceConfiguration`'s property after posting, since the config's `handleEvent:` runs on the same serial queue (already done in `TeakDeviceConfigurationTests`, though that test calls `handleEvent:` directly and bypasses the queue entirely, which is fine for its scope).

Not blocking — the current approach works and the comment explains it — but if `TeakEvent`'s queue ever becomes concurrent or gets renamed, these tests break silently.

**5. Test — `TeakDeviceConfigurationTests` calls `handleEvent:` directly.** That's consistent with unit-test scope (isolates configuration state transitions from the event bus), but it means the end-to-end path `[Teak registerPushToStartToken:]` → `PushRegistrationEvent` → `TeakEvent` dispatch → `TeakDeviceConfiguration.handleEvent:` → `liveActivityPushToStartToken` is not covered by any single test. If something wires up wrong between `Teak.m` and the config object, the unit tests would still pass. Low risk given how faithfully the pattern mirrors `pushToken`, but worth knowing.

**6. Docstring wording nit.** `Teak.h:601-602`: "Safe to call on any iOS version: on versions below iOS 17.2 the OS never produces a push-to-start token, so this method simply won't be invoked." The second clause is true from the SDK's perspective but reads oddly as a guarantee — it's really a statement about caller behavior, not SDK behavior. Consider: "Safe to call on any iOS version. On iOS versions below 17.2 the OS never produces a push-to-start token, so in practice callers will only invoke this on iOS 17.2+."

### Risks
- **Low overall.** Additive change, guarded by `enablePushKey`, non-empty check before serialization, isolated to a new event-type case.
- **Verify server-side contract:** PR description says C-606 is already in place server-side accepting `live_activity_push_to_start_key`. Worth a manual confirmation (curl to staging or a cleanroom run) before merging to 4.3-stable, since the PR description leaves that step unchecked.

### Verdict
Approvable with minor suggestions. Items 1, 2, 6 are easy in-PR tweaks; items 3, 4, 5 are worth considering but could be follow-ups. The manual verification step in the PR's own test plan is the main gap before merging.

AlexSc:
Addressed review items 1, 2, and 6 in a004987:

- **1.** `registerPushToStartToken:` now logs `"token cannot be nil"` vs `"token cannot be empty"` as two distinct paths.
- **2.** Added a comment in `TeakDeviceConfiguration.m` next to the `LiveActivityPushToStartRegistered` case explaining why there's no symmetric unregister path (OS rotates the token in place; it doesn't revoke it).
- **6.** Reworded the iOS 17.2 availability sentence so it reads as a statement about caller behavior rather than an SDK guarantee about OS behavior.

Thoughts on the remaining items:

- **3. `dispatchUserDataEvent` gap.** Intentionally out of scope — `UserDataEvent` is a game-facing echo via `TeakUserData`, and the game is the one that just passed the token to Teak, so it doesn't need it echoed back. C-608's acceptance criterion is server-side only (token appears in `push_keys` after next session report). Happy to revisit if a downstream consumer actually needs it.
- **4. Test private-API coupling.** Aware of the trade-off. The `dispatch_sync` barrier gives deterministic "did not post" assertions; an `XCTestExpectation` with a timeout would introduce either flakiness (short timeout, rare false-positives) or slow tests (long timeout). The category is test-only and the event queue's serial nature is a load-bearing property of the production code — if it ever becomes concurrent, both the test barrier and real handlers would need revisiting. Can revisit if we'd rather accept the fuzziness.
- **5. End-to-end test gap.** Acknowledged. Mitigated in practice by how closely the new code mirrors the `pushToken` pattern (same event-class, same device-config shape, same KVO wiring), but happy to add a dedicated end-to-end test in a follow-up if the mirroring is uncomfortable as test cover.

Also added a manual verification note to the PR body — planning to wire this into the swift-cleanroom and confirm `live_activity_push_to_start_key` lands server-side before taking it out of draft.

AlexSc:
Can we get push to start tokens implicitly? All live activities share a single pts token, so this feels like something we could work into a lifecycle hook or swizzle.

Teak/Configuration/TeakDeviceConfiguration.m
```diff
@@ -185,6 +188,16 @@ - (void)handleEvent:(TeakEvent* _Nonnull)event {
     case PushUnRegistered: {
       if (self.pushToken != nil) self.pushToken = nil;
     } break;
+    case LiveActivityPushToStartRegistered: {
+      // No symmetric "unregister" case: push-to-start tokens are per-device and persistent.
+      // The OS rotates the token in place via Activity.pushToStartTokenUpdates; it doesn't
```
AlexSc: Are we sure about this? How do we test that assumption, and what are the consequences of being wrong?

Teak/Teak.h
```diff
@@ -590,6 +590,21 @@ typedef void (^TeakLogListener)(NSString* _Nonnull event,
  */
 + (nonnull TeakOperation*)startedLiveActivity:(nonnull NSString*)activityId withToken:(nonnull NSData*)pushToken systemActivityId:(nonnull NSString*)systemActivityId;
 
+/**
+ * Report the live activity push-to-start token for this device.
+ *
+ * Call this once at launch, after observing ``Activity<Attributes>.pushToStartTokenUpdates``, and
+ * whenever the OS rotates the token. The token is per-device and applies to all push-to-start
```
AlexSc: We don’t need to document internal implementation details

AlexSc:
Short answer: not cleanly, because Apple's push-to-start API is per-`ActivityAttributes`-type and the core SDK is Obj-C today.

**Structural obstacle**
- `Activity<T>.pushToStartTokenUpdates` is a Swift `AsyncSequence` parameterized on a Swift-only protocol. Obj-C can't name or observe those generic types.
- There's no device-wide hook analogous to `didRegisterForRemoteNotificationsWithDeviceToken:` for push-to-start — ActivityKit only exposes the typed AsyncSequence.
- Swizzling isn't viable — there's no `NSObject`-backed method we could intercept.

**Three options**

1. **Status quo** — keep `registerPushToStartToken:`. Game writes its own 4-line \`for await\` loop. Explicit, portable, zero Swift dep.
2. **Swift helper in Teak** — add a Swift module with \`Teak.observePushToStartTokens(for: MyActivityAttributes.self)\` that runs the \`for await\` loop and calls through to the Obj-C API. Reduces game boilerplate to one line per attributes type but still needs the game to name its types. Costs a \`Teak.swift\` target / new subspec.
3. **Sentinel `ActivityAttributes` inside Teak** — if the \"one token per device across all types\" claim in C-608 is literally true at the iOS level, Teak could define its own internal \`_TeakPushToStartObserver: ActivityAttributes\` type and observe *its* \`pushToStartTokenUpdates\`. Fully implicit from the game's perspective, but load-bearing on iOS delivering a push-to-start token for an \`ActivityAttributes\` type that has never been started as an activity. I'm not confident that holds — would want to validate on-device before relying on it. Also adds a Swift target.

Filing a follow-up Linear to spike option 3 — if the sentinel receives tokens, that's the ergonomics win we want. If it doesn't, option 2 becomes the fallback. Either way, keeping C-608 as-is as the primitive.

Teak/Configuration/TeakDeviceConfiguration.m
AlexSc: Fair point — I was extrapolating from C-606's 'per-device and persistent' wording and don't actually know whether ActivityKit surfaces a revocation signal. Dropped the claim in 72bee0b and filed [C-664](https://linear.app/teakio/issue/C-664/sdk-verify-activitykit-push-to-start-token-revocation-behavior) to empirically verify what `pushToStartTokenUpdates` does under Live Activities being disabled, notification revocation, reinstall, and device restore. If the AsyncSequence does surface revocation, C-664's deliverable is wiring up a `PushToStartUnRegistered`-equivalent event.

Teak/Teak.h
AlexSc: Agreed. 72bee0b drops the per-device/stores-on-config prose; the docstring now just describes how the caller uses the API (observe the AsyncSequence, forward each token) and leaves Apple's token semantics + Teak's internals out of it.

Teak/Teak.h
```diff
@@ -590,6 +590,20 @@ typedef void (^TeakLogListener)(NSString* _Nonnull event,
...
 + (nonnull TeakOperation*)startedLiveActivity:(nonnull NSString*)activityId withToken:(nonnull NSData*)pushToken systemActivityId:(nonnull NSString*)systemActivityId;
 
+/**
+ * Report a live activity push-to-start token to Teak.
+ *
+ * Call this from any location where your app observes
+ * ``Activity<Attributes>.pushToStartTokenUpdates`` — typically at launch and again
+ * whenever the OS emits a rotated token.
+ *
+ * Safe to call on any iOS version. Push-to-start tokens are an iOS 17.2+ feature,
```
AlexSc: Unnecessary

AlexSc: Agreed, dropped in 2217cff. The availability note was restating what callers already know from the API they're observing.

AlexSc:
Linear: [C-608 — SDK: push-to-start token registration](https://linear.app/teakio/issue/C-608/sdk-push-to-start-token-registration)

AlexSc:
Follow-ups filed from this review:

- [C-663 — SDK: implicit push-to-start token observation (spike + implement)](https://linear.app/teakio/issue/C-663/sdk-implicit-push-to-start-token-observation-spike-implement) — spike the sentinel `ActivityAttributes` approach, then implement either that or the Swift-helper fallback.
- [C-664 — SDK: verify ActivityKit push-to-start token revocation behavior](https://linear.app/teakio/issue/C-664/sdk-verify-activitykit-push-to-start-token-revocation-behavior) — empirically test what `pushToStartTokenUpdates` does under Live Activities off / notifications revoked / reinstall / restore, wire up a clear path if any of those surface revocation.

PR #124, opened 2026-04-18
feat(C-644): Attribute Live Activity taps via TeakLiveActivityLaunchData

AlexSc:
## Summary
- Hook iOS's activity-resumption callback (`NSUserActivityTypeLiveActivity`, userInfo key `WGWidgetUserInfoKeyActivityID`) to construct a new `TeakLiveActivityLaunchData` — a sibling of `TeakNotificationLaunchData` under `TeakAttributedLaunchData`.
- `sessionAttribution` adds `teak_live_activity_id` (value = Apple's `Activity.id`), so `TeakSession`'s existing `launchDataOperation.result sessionAttribution` merge carries it into the identify-user POST with no session-side changes.
- Browsing-web/universal-link attribution path is unchanged: both forks now dispatch through `+[TeakLaunchDataOperation fromUserActivity:]`.

## Key decisions
- **Detection pivot is Apple's own constants, not a Teak convention.** The pods cleanroom's device logs show iOS delivers `activityType == NSUserActivityTypeLiveActivity` with `userInfo[@"WGWidgetUserInfoKeyActivityID"] = Activity.id` — we match those string values verbatim rather than requiring the game to stuff a Teak-specific key into its widget userInfo.
- **No ActivityKit/WidgetKit link.** The constants are declared as local `NSString* const` literals so the SDK's iOS 11 deployment target isn't disturbed.
- **`to_h` is a full override (skips the attributed chain).** LA taps carry no schedule/creative/reward/channel/deep-link metadata, and the attributed `to_h` calls `TeakLink_WillHandleDeepLink` which requires `TeakConfiguration` — the minimal override is both semantically correct for LA and testable without SDK init.
- **Single dispatcher point.** Both the scene-hooks `willConnectToSession` cold-start path and the scene/app-delegate `continueUserActivity:` tap-while-running path already funnel into `application:continueUserActivity:restorationHandler:`, so one call site covers both.

## Test plan
New `TeakLiveActivityLaunchDataTests.m` (48 total tests pass — `bundle exec fastlane test`):
- [x] Class shape: inherits from `TeakAttributedLaunchData`, init populates `systemActivityId`.
- [x] `sessionAttribution` contains `teak_live_activity_id` mapped to the system activity id and does not carry a spurious `teak_notif_id`.
- [x] `to_h` contains `teakSystemActivityId`.
- [x] `+fromLiveActivityTap:` factory: the operation runs and produces a `TeakLiveActivityLaunchData` whose `sessionAttribution` carries `teak_live_activity_id` — this is the mock-test path the ticket calls out (identify-user POST payload includes `teak_live_activity_id`).
- [x] `+fromUserActivity:` dispatcher: recognizes a simulated LA-tap `NSUserActivity`, ignores LA taps with nil/empty system activity id, ignores unknown activity types, and still routes browsing-web activities to the universal-link path (no regression).

Live-device verification is out of scope here — it's the C-646 cleanroom ticket this PR unblocks.

Closes C-644.

### DIFF
--- a/Teak/Core/TeakLaunchData.m
+++ b/Teak/Core/TeakLaunchData.m
@@ -89,6 +89,7 @@ static NSString* const kTeakWGWidgetUserInfoKeyActivityID = @"WGWidgetUserInfoKe
 }
 
 + (TeakLaunchDataOperation*)fromLiveActivityTap:(NSString*)systemActivityId {
+  TeakLog_i(@"live_activity.attribution.received", @{@"systemActivityId" : systemActivityId});
   TeakLiveActivityLaunchData* launchData = [[TeakLiveActivityLaunchData alloc] initWithSystemActivityId:systemActivityId];
   return [[TeakLaunchDataOperation alloc] initWithLaunchData:launchData];
 }
--- a/Teak/Core/TeakLaunchData.m
+++ b/Teak/Core/TeakLaunchData.m
@@ -41,6 +41,18 @@ extern BOOL TeakLink_WillHandleDeepLink(NSURL* deepLink);
 - (id)initWithUrl:(NSURL*)url andShortLink:(NSURL*)shortLink;
 @end
 
+@interface TeakLiveActivityLaunchData ()
+@property (copy, nonatomic, readwrite) NSString* systemActivityId;
+
+- (id)initWithSystemActivityId:(NSString*)systemActivityId;
+@end
+
+// Activity-resumption detection constants. These match the string values ActivityKit
+// (iOS 16.2+) and WidgetKit use when iOS delivers a Live Activity tap. Declared as
+// literals so the SDK doesn't have to link those frameworks on its iOS 11 deployment target.
+static NSString* const kTeakNSUserActivityTypeLiveActivity = @"NSUserActivityTypeLiveActivity";
+static NSString* const kTeakWGWidgetUserInfoKeyActivityID = @"WGWidgetUserInfoKeyActivityID";
+
 /// Implementations
 
 #define NewIfNotOld(x, y) (x == nil ? y : x)
@@ -74,6 +86,28 @@ extern BOOL TeakLink_WillHandleDeepLink(NSURL* deepLink);
 + (TeakLaunchDataOperation*)fromOpenUrl:(NSURL*)url {
   TeakLaunchData* launchData = [TeakLaunchDataOperation launchDataFromUrl:url withShortlink:nil];
   return [[TeakLaunchDataOperation alloc] initWithLaunchData:launchData];
+}
+
++ (TeakLaunchDataOperation*)fromLiveActivityTap:(NSString*)systemActivityId {
+  TeakLiveActivityLaunchData* launchData = [[TeakLiveActivityLaunchData alloc] initWithSystemActivityId:systemActivityId];
+  return [[TeakLaunchDataOperation alloc] initWithLaunchData:launchData];
+}
+
++ (TeakLaunchDataOperation*)fromUserActivity:(NSUserActivity*)userActivity {
+  if (userActivity == nil) return nil;
+
+  if ([userActivity.activityType isEqualToString:NSUserActivityTypeBrowsingWeb]) {
+    return [TeakLaunchDataOperation fromUniversalLink:userActivity.webpageURL];
+  }
+
+  if ([userActivity.activityType isEqualToString:kTeakNSUserActivityTypeLiveActivity]) {
+    NSString* systemActivityId = userActivity.userInfo[kTeakWGWidgetUserInfoKeyActivityID];
+    if ([systemActivityId isKindOfClass:[NSString class]] && systemActivityId.length > 0) {
+      return [TeakLaunchDataOperation fromLiveActivityTap:systemActivityId];
+    }
+  }
+
+  return nil;
 }
 
 + (TeakLaunchDataOperation*)unattributed {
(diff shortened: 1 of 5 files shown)


Teak/Core/TeakLaunchData.m
```diff
@@ -76,6 +88,29 @@ + (TeakLaunchDataOperation*)fromOpenUrl:(NSURL*)url {
   return [[TeakLaunchDataOperation alloc] initWithLaunchData:launchData];
 }
 
++ (TeakLaunchDataOperation*)fromLiveActivityTap:(NSString*)systemActivityId {
+  TeakLog_i(@"live_activity.attribution.received", @{@"systemActivityId" : systemActivityId});
```
AlexSc, line 92: Will the session attribution get reported to the server? If so, how?

AlexSc: Yes — via the existing identify-user POST. The trace:

1. This factory builds a `TeakLaunchDataOperation` wrapping a `TeakLiveActivityLaunchData`.
2. `Teak.m application:continueUserActivity:` hands it to `[TeakSession didLaunchWithData:]`.
3. `TeakSession.m:583 didLaunchWithData:` — if a session is already active (e.g. `Expiring` after a background trip), it creates a new session via `initWithSession:` that inherits `userId`/`facebookAccessToken`, expires the old one, and sets `self.launchDataOperation` on the new session.
4. The new session walks `Allocated → Created → Configured` (`Created` in `init` at L395, `Configured` once `TeakRemoteConfiguration` loads).
5. `TeakSession.m:749` KVO on `currentState → Configured` enqueues `identifyUserOperation`, which adds `self.launchDataOperation` as an NSOperation dependency (L441) so the launch data is guaranteed finished before the request fires.
6. `sendUserIdentifier` (L186) transitions to `IdentifyingUser`, builds the `/games/{appId}/users.json` payload, and at **L269** does `[payload addEntriesFromDictionary:[self.launchDataOperation.result sessionAttribution]]` — the LA override returns `{teak_live_activity_id: <Activity.id>}`, which lands in the POST.
7. Backend resolves the attribution via C-639.

No C-644-specific network call; the existing session-identify path carries it through.

One consequence worth calling out: because `didLaunchWithData:` is the same entry point notifications/universal links use, an LA tap on a backgrounded app rotates the session and re-sends users.json. That matches how a notification tap or universal link behaves today.

AlexSc: Can the server reply with enhanced/enriched attribution data for our post launch summary?

AlexSc: Yes — via the users.json reply's `deep_link` field. Tracing the current code plus what I just pushed in `c584b1b`:

**Enrichment plumbing (pre-existing):** When the backend resolves attribution and returns `deep_link` in the `/games/{appId}/users.json` reply, TeakSession.m:295-303 calls `[launchDataOperation updateDeepLink:enrichedURL withLaunchLink:launchLink]`. For attributed launch data that dispatches to `TeakAttributedLaunchData.updateDeepLink:`, which rebuilds scheduleName / scheduleId / creativeName / creativeId / rewardId / channelName / deepLink / optOutCategory via `NewIfNotOld(old, new)` — "keep what we already had, else take from the enriched URL." Then `processAttributionAndDispatchEvents` (TeakSession.m:467-471) posts `TeakPostLaunchSummary` with `[launchData to_h]`, so observers see the enriched fields.

**Two things that were in the way:**

1. **My `to_h` override was bypassing the attributed chain** (emitted only `launch_link` + `teakSystemActivityId`). So even if enrichment landed on the instance, my override never surfaced it to `TeakPostLaunchSummary`. Switched to call super, matching `TeakNotificationLaunchData`.

2. **`TeakAttributedLaunchData.initWithAttributedLaunchData:andUpdatedDeepLink:` was building `newLaunchData` via the parent `initWithUrl:`** (not `initWithUrl:andShortLink:`). The parent's `initWithUrl:` doesn't parse teak_* params, so `newLaunchData.scheduleId` etc. were always nil, meaning `NewIfNotOld(old, nil)` always returned old — the enrichment merge was effectively a no-op for every slot. This was pre-existing and affected every attributed launch data type. Fix is one line. It's behavior-preserving for notifications/URL launches (they arrive with `old` already non-nil, so `NewIfNotOld` still returns old), and unlocks enrichment for LA (where old is nil). I bundled it into this PR because without it the LA post-launch summary is stuck with just `teakSystemActivityId` even when the server replies with a full attributed deep link — and verifying the bundled change is what the new `testUpdateDeepLinkEnrichesAttributedFieldsAndPreservesSystemActivityId` test does end-to-end (sessionAttribution AND to_h both carry enriched teak_schedule_id/teak_creative_id/teak_reward_id + systemActivityId preserved).

Let me know if you'd rather see #2 broken out into its own PR.

AlexSc: Yea, break that out into its own PR please. We'll get that merged first.

AlexSc: Filed [C-672](https://linear.app/teakio/issue/C-672/userscontroller-return-schedule-level-deep-link-in-usersjson-reply-on) for the server side — UsersController extension so the users.json reply carries a `deep_link` with `teak_schedule_id` (+ optional `teak_schedule_name`) when `teak_live_activity_id` resolves. Low priority, Backlog, relates back to C-639 and C-644. iOS side is already ready to consume the enriched reply once that lands (no further SDK work needed thanks to #125 + the `to_h → super` swap in this PR).

PR #131, opened 2026-04-23
Fix NSInvalidArgumentException on non-JSON error responses

AlexSc:
## Summary
- A 500 response with a non-JSON body (e.g. an HTML error page) caused `NSJSONSerialization` to return nil, which `URLSession:task:didCompleteWithError:` forwarded downstream as the reply payload; reply parsers like the one for `/me/live_activities` then crashed in `@{@"response" : reply}` with `NSInvalidArgumentException: attempt to insert nil object from objects[0]`.
- Root cause fixed at the parse step: extracted `+[TeakRequest parseJSONResponseData:]` which always returns a dictionary (nil/empty/non-JSON/non-object bodies all collapse to `@{}`).
- Also fixes a latent, inverted ternary for `report_client_error` handling that returned nil for `title` when the key was absent (and the literal string "Configuration Error" when the key was present) — extracted `+[TeakRequest titleForClientError:]` with correct fallback and NSString type-check.

## Key decisions
- Guarded at the source in `TeakRequest` rather than sprinkling null-checks into each reply parser. Future reply parsers inherit the guarantee instead of being one bug away from the same crash.
- Non-object JSON (top-level arrays/strings/numbers) also collapses to `@{}`. The previous code cast the parse result to `NSDictionary*` without checking, so those shapes would have crashed the same reply parsers; the helper closes that class too.
- Title helper rejects NSNull as well as missing keys, so a server response of `"title": null` is handled the same as an absent key.
- Based off `4.3-stable` because the reported crash is on SDK `4.3.11-rc0-45-g5bf8484`.

## Test plan
- [x] `bundle exec fastlane test` — 103 unit tests pass, including 8 new in `TeakRequestResponseParsingTests` and 5 new in `TeakRequestClientErrorTests`.
- [x] New nil-payload tests cover: nil data, zero-length data, HTML error body, garbage bytes, top-level JSON array, top-level JSON string, valid JSON object, empty JSON object.
- [x] New title-ternary tests cover: title present, title missing, nil dict, empty dict, NSNull title.
- [ ] Manual: reproduce the original 500-from-`/me/live_activities` scenario and confirm no crash (reporter to verify in the environment that produced the original Sentry event).

### DIFF
--- a/Automated/AutomatedTests/TeakRequestClientErrorTests.m
+++ b/Automated/AutomatedTests/TeakRequestClientErrorTests.m
@@ -24,10 +24,7 @@
 
 - (void)testFallsBackToConfigurationErrorWhenTitleMissing {
   NSDictionary* err = @{@"message" : @"Something went wrong"};
-  XCTAssertEqualObjects([TeakRequest titleForClientError:err], @"Configuration Error",
-                        @"Title must fall back to 'Configuration Error' when the key is missing — "
-                        @"the original ternary was inverted and returned nil here, crashing "
-                        @"TeakIntegrationChecker.reportError:forCategory:");
+  XCTAssertEqualObjects([TeakRequest titleForClientError:err], @"Configuration Error");
 }
 
 - (void)testFallsBackToConfigurationErrorForNilDict {
@@ -40,9 +37,7 @@
 
 - (void)testFallsBackToConfigurationErrorWhenTitleIsNSNull {
   NSDictionary* err = @{@"title" : [NSNull null]};
-  XCTAssertEqualObjects([TeakRequest titleForClientError:err], @"Configuration Error",
-                        @"NSNull must be treated the same as a missing key — it cannot be used as "
-                        @"a dictionary key by the downstream integration checker.");
+  XCTAssertEqualObjects([TeakRequest titleForClientError:err], @"Configuration Error");
 }
 
 @end
--- a/Automated/AutomatedTests/TeakRequestClientErrorTests.m
+++ b/Automated/AutomatedTests/TeakRequestClientErrorTests.m
@@ -1,0 +1,48 @@
+#import <XCTest/XCTest.h>
+
+#import "TeakRequest.h"
+#import <Teak/Teak.h>
+
+@import OCHamcrest;
+@import OCMockito;
+
+// Re-expose the internal helper so tests can exercise it without
+// driving a full request lifecycle. See CLAUDE.md "Header imports in tests".
+@interface TeakRequest (ClientErrorTests)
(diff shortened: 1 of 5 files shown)


Automated/AutomatedTests/TeakRequestClientErrorTests.m
```diff
@@ -0,0 +1,48 @@
...
+
+- (void)testReturnsTitleFromDictWhenPresent {
+  NSDictionary* err = @{@"title" : @"Bad API Key", @"message" : @"Invalid key"};
+  XCTAssertEqualObjects([TeakRequest titleForClientError:err], @"Bad API Key");
+}
+
+- (void)testFallsBackToConfigurationErrorWhenTitleMissing {
+  NSDictionary* err = @{@"message" : @"Something went wrong"};
+  XCTAssertEqualObjects([TeakRequest titleForClientError:err], @"Configuration Error",
+                        @"Title must fall back to 'Configuration Error' when the key is missing — "
```
AlexSc: Unnecessary.

Teak/TeakRequest+Internal.h
```diff
@@ -16,4 +16,15 @@
 @property (strong, nonatomic, readwrite) NSString* _Nonnull method;
 
 - (nullable TeakRequest*)initWithSession:(nonnull TeakSession*)session forHostname:(nonnull NSString*)hostname withEndpoint:(nonnull NSString*)endpoint withPayload:(nonnull NSDictionary*)payload method:(nonnull NSString*)method callback:(nullable TeakRequestResponse)callback addCommonPayload:(BOOL)addCommonToPayload;
+
+// Parse an HTTP response body into a reply dictionary. Always returns a
+// dictionary — nil/empty/non-JSON/non-object bodies all collapse to @{} so
+// downstream code (reply parsers, dictionary literals) never sees nil.
++ (NSDictionary* _Nonnull)parseJSONResponseData:(NSData* _Nullable)data;
```
AlexSc: I feel like it would be a good idea to propagate errors out of this for Sentry logging in the future

Automated/AutomatedTests/TeakRequestClientErrorTests.m
AlexSc: Dropped in f5ecf9b — pruned the same anti-pattern from the other 3 verbose assertion descriptions across both new test files too.

Teak/TeakRequest+Internal.h
AlexSc: Done in ad9c661 — added `error:` out-parameter, synthesized a descriptive NSError for the "parsed but not a dictionary" case as well, and wired the URLSession delegate to `TeakLog_e(@"request.reply.parse_error", parseError)` so Sentry will get these.

PR #134, opened 2026-04-27
C-697: iOS claim_mode opt-in, click payload, enum extension

TeakBuilds:
Closes https://linear.app/teakio/issue/C-697/ios-sdk-claim-mode-opt-in-click-payload-enum-extension

## Summary

SDK A "skeleton" for Simplify Rewards on iOS: declares the app's claim mode, wires it onto every reward click, and prepares the reward-status enum for the new server outcomes. Self-contained — no click-time JWT handling here (that's C-698).

- `TeakAppConfiguration.claimMode` reads `TeakClaimMode` from Info.plist, defaulting to `@"legacy"`. Surfaced in `to_h` / `description` for the `configuration.app` log event.
- `TeakReward` click POST always carries `claim_mode = appConfiguration.claimMode` — explicit on every request, including `claim_mode=legacy` for non-opted-in apps.
- `TeakRewardStatus` enum gains `kTeakRewardStatusPlayerIneligible` (7), `kTeakRewardStatusNoRewardAvailable` (8), `kTeakRewardStatusClaimModeUnsupported` (9), with matching string-to-enum entries for `player_ineligible`, `no_reward_available`, `claim_mode_unsupported`.
- After the settings.json reply lands, `TeakRemoteConfiguration` validates the configured mode against the returned `supported_claim_modes` and emits a `claim_mode.unsupported` warning (new `TeakLog_w` / WARN level) when there's a mismatch. Warning only — never an init abort.

## Key decisions

- The status-string mapping is hoisted into `+[TeakReward rewardStatusForString:]` and the validation logic into `+[TeakRemoteConfiguration validateClaimMode:againstSupported:]`. Both are private surfaces re-declared in test class extensions, avoiding the need to drag `TeakSession` or HTTP plumbing into unit tests.
- `TeakLog_w` (WARN) is added alongside the existing `TeakLog_i` / `TeakLog_e` overload sets so the new init-time validation message has its own level.
- Init-time validation hooks into the existing settings.json reply callback rather than subscribing to `RemoteConfigurationReady` as a separate observer — `session.appConfiguration` is already in scope there.

## Test plan

- `bundle exec fastlane test` → 129 tests, 0 failures.
- New coverage:
  - `TeakRewardStatusMappingTests` — five tests: existing statuses, each of the three new statuses (with explicit raw-value assertions for 7/8/9), unknown string falls through to `kTeakRewardStatusUnknown`.
  - `TeakAppConfigurationClaimModeTests` — `claimMode` defaults to `@"legacy"` when Info.plist has no `TeakClaimMode`; `to_h` includes the field.
  - `TeakRemoteConfigurationClaimModeTests` — return-value behavior across supported/unsupported/missing/non-array inputs, plus a `logListener`-based check that the warning fires at WARN level with the configured/supported pair in `event_data` and stays silent when the mode is supported.

Did not run `./format-code` (per the worktree CLAUDE.md, the script is currently broken; followed the existing style by hand).

## Readers left behind

- `kTeakRewardStatusClaimModeUnsupported` is wire-ready but only meaningful once the server can return `claim_mode_unsupported` — that's a Carrot-side change and out of scope here.
- Click-time JWT minting / `event=click_*` emission is C-698 / SDK B; this PR deliberately stops at the wire.

### DIFF
--- a/Teak/Configuration/TeakAppConfiguration.m
+++ b/Teak/Configuration/TeakAppConfiguration.m
@@ -22,6 +22,7 @@ BOOL Teak_isProductionBuild(void) {
 @property (strong, nonatomic, readwrite) NSString* bundleId;
 @property (strong, nonatomic, readwrite) NSString* appVersion;
 @property (strong, nonatomic, readwrite) NSString* _Nonnull appVersionName;
+@property (strong, nonatomic, readwrite) NSString* claimMode;
 @property (strong, nonatomic, readwrite) NSSet* urlSchemes;
 @property (nonatomic, readwrite) BOOL isProduction;
 @property (nonatomic, readwrite) BOOL traceLog;
@@ -68,6 +69,15 @@ BOOL Teak_isProductionBuild(void) {
     }
     teak_catch_report;
 
+    self.claimMode = @"legacy";
+    teak_try {
+      NSString* claimMode = [[NSBundle mainBundle] infoDictionary][@"TeakClaimMode"];
+      if ([claimMode isKindOfClass:[NSString class]] && claimMode.length > 0) {
+        self.claimMode = claimMode;
+      }
+    }
+    teak_catch_report;
+
 #define kTeakLogTrace @"TeakLogTrace"
 #define kTeakSDK5Behaviors @"TeakSDK5Behaviors"
 #define kTeakDoNotRefreshPushToken @"TeakDoNotRefreshPushToken"
@@ -89,6 +99,7 @@ BOOL Teak_isProductionBuild(void) {
     @"apiKey" : self.apiKey,
     @"bundleId" : self.bundleId,
     @"appVersion" : self.appVersion,
+    @"claimMode" : self.claimMode,
     @"isProduction" : self.isProduction ? @YES : @NO,
     @"traceLog" : self.traceLog ? @YES : @NO,
     @"sdk5Behaviors" : self.sdk5Behaviors ? @YES : @NO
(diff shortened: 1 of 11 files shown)


Teak/Configuration/TeakAppConfiguration.m
```diff
@@ -68,6 +69,15 @@ - (id)initWithAppId:(nonnull NSString*)appId apiKey:(nonnull NSString*)apiKey {
     }
     teak_catch_report;
 
+    self.claimMode = @"legacy";
+    teak_try {
+      NSString* claimMode = [[NSBundle mainBundle] infoDictionary][@"TeakClaimMode"];
```
AlexSc: Let's name the parameter TeakRewardClaimMode

AlexSc (changes requested):

Teak/Configuration/TeakAppConfiguration.m
TeakBuilds: [worker-697] Done in 5e53434. Renamed the Info.plist key `TeakClaimMode` -> `TeakRewardClaimMode` here (line 74), and cascaded:

- `TeakRemoteConfiguration.m` — the WARN message text now says `TeakRewardClaimMode` so misconfigured apps see the current public name in the log.
- `TeakAppConfigurationClaimModeTests.m` — assertion message + comment updated. Default-to-legacy behavior is unchanged so no logic-side test edits needed.

Left the internal `TeakAppConfiguration.claimMode` property as-is — your comment was anchored on the literal plist-key string, and renaming the property would cascade through every `claimMode` reference (the click-payload site in `TeakReward.m`, `+validateClaimMode:againstSupported:`, `to_h`, `description`, the `event_data` key in the WARN, and 5 tests) without changing the public surface. Happy to do that rename too if you want it; let me know.

Note for cross-SDK awareness: Android (PR #144) uses `io_teak_claim_mode` (no "reward") for its resource name. After this rename iOS and Android diverge slightly on the developer-facing key, which I think is fine because the cross-SDK contract is the wire payload + the WARN `event_data` shape (already aligned), not the client-side config key. Flagging in case you'd rather both platforms scope to reward.

`bundle exec fastlane test` → 129 / 0.

AlexSc: productops: please align android to include reward

AlexSc (approved):

Teak/Configuration/TeakAppConfiguration.m
TeakBuilds: [worker-700] Android alignment in flight: pushed `io_teak_claim_mode` -> `io_teak_reward_claim_mode` to PR #144 (https://github.com/GoCarrot/teak-android/pull/144) in commit 71732c61. Resource string + changelog + test comment updated; in-memory `claimMode` field unchanged. Idling for CI + re-review.

Issue #10, opened 2020-01-14
Track foreground notification receipts


AlexSc:
If a notification is received when the game is in the foreground the call to /notification_received should additionally include notification_placement=foreground

Because this comes from the notification service it might not be doable. We should first determine which is fired first -- the notification service or the in-app delegate.

---

Repo: [`teak-ios`](<https://github.com/GoCarrot/teak-ios>)

PR #135, opened 2026-04-27
[worker-698] iOS SDK: JWT-mode click-time events + poll + ack

TeakBuilds:
Closes https://linear.app/teakio/issue/C-698

## Summary

Implements the iOS SDK's click-time side of the JWT-mode reward surface. Adds three new `NSNotification` names (`TeakOnRewardJwtIssued`, `TeakOnRewardClaimPending`, `TeakOnRewardClaimResolved`); branches the click reply on its wire `status` field; mints `session_attribution` from `launchData.to_h` at click-request-build time; runs a click-time `NSTimer` exponential-backoff poll against `/claim_status` for `claim_pending` claims; fires the resolved event and POSTs `/claim_ack` on terminal status. Does NOT include the session-start sweep (that's C-699, blocked by this).

The PR also absorbs the iOS-side spec parity work from `taro/docs/session_attribution_spec.md`: `TeakLaunchData.to_h` now emits the canonical eleven-key wire shape (drift item 1, all-keys-always-present); `teakDeepLink` surfaces the resolved deep link instead of the outer launch URL (drift item 2); `teakOptOutCategory` defaults to the literal "teak" on attributed launches (drift item 3), with the `TeakLiveActivityLaunchData` override returning to NSNull per the LA fixture.

## Key decisions

- **In-session optimization for resolved events.** The `TeakOnRewardClaimResolved` userInfo merges the polled reply with `launchData.to_h` (the SDK's in-memory launch-data state) instead of consuming the server-persisted `session_attribution` blob. Per the spec, the click-time poll fires within the same session as the click, so the SDK still has the launch-data state and can skip the round-trip. C-699's session-start sweep is what consumes the persisted blob for cross-session resolutions.
- **Status branch is conservative on unknowns.** Unknown wire statuses fall back to `TeakOnReward` rather than fanning out to a new event, matching how host games already handle the legacy reward enum's `unknown` value. Forward-compatible: future enum values render as legacy events that existing host code already routes.
- **TDD discipline as two commits.** Commit 1 adds the tests + production stubs (compiles, 83 tests fail at runtime). Commit 2 swaps the stubs for the real implementation (all 154 tests pass). Reviewer can read the contract before the implementation.

## Test plan

- [x] `bundle exec fastlane test` → 154 tests, 0 failures (was 129; +25 new tests across `SessionAttributionSpecParityTests`, `JwtModeRewardEventsTests`, `ClaimStatusPollTests`).
- [x] Spec-parity tests round-trip the canonical wire shape for unattributed / rewardlink / notification / live-activity launches.
- [x] Status-branch tests cover `grant_reward` / `token_issued` / `claim_pending` / `claim_mode_unsupported` / unknown.
- [x] Backoff tests cover initial-delay / doubling / ceiling-clamp at multiple base values.
- [ ] Manual verification against a live taro `claim_pending` flow — gated on the rest of the simplify-rewards integration landing (C-699 / Carrot side / ARB integration).

## Readers left behind

Nothing intentionally deferred from C-698's scope.

### DIFF
--- a/Teak/Teak.h
+++ b/Teak/Teak.h
@@ -18,8 +18,42 @@ extern NSString* _Nonnull const TeakNotificationAppLaunch;
  */
 extern NSString* _Nonnull const TeakOnReward;
 
+/**
+ * Use this named notification to listen for the JWT-issued response when an
+ * app is configured for client-side JWT reward claims.
+ *
+ * The notification's userInfo carries the JWT token, event id, reward,
+ * claim source, teak reward id, and the launch-data attribution fields so
+ * the host game can render reward UI and forward the JWT to its own
+ * backend.
+ */
 extern NSString* _Nonnull const TeakOnRewardJwtIssued;
+
+/**
+ * Use this named notification to listen for the pending-claim response when
+ * an app is configured for server-side JWT reward claims.
+ *
+ * The notification's userInfo carries the event id (used by the SDK's
+ * click-time poll loop) plus the reward and attribution context so the host
+ * game can render an optimistic surface while the claim is in flight.
+ */
 extern NSString* _Nonnull const TeakOnRewardClaimPending;
+
+/**
+ * Use this named notification to listen for the resolved-claim response from
+ * the SDK's click-time poll loop. Fires when the server reaches a terminal
+ * state (completed or failed) for a previously-pending claim.
+ *
+ * The notification's userInfo carries the polled status, the customer's
+ * server response, and the launch-data attribution fields.
+ *
+ * Delivery is at-least-once: if the SDK's `/claim_ack` post fails (network
+ * blip, app backgrounded mid-flight), the server's session-start sweep will
+ * resurface the same resolved claim on the next launch and fire this
+ * notification again with the same `event_id`. Host games SHOULD idempotent-
+ * key any reward grant on the `event_id` field of the userInfo dictionary
+ * so a duplicate delivery doesn't grant the reward twice.
+ */
 extern NSString* _Nonnull const TeakOnRewardClaimResolved;
 
 /**
--- a/Teak/Teak.h
+++ b/Teak/Teak.h
@@ -26,6 +26,11 @@ extern NSString* _Nonnull const TeakOnReward;
  * claim source, teak reward id, and the launch-data attribution fields so
  * the host game can render reward UI and forward the JWT to its own
  * backend.
+ *
+ * Access the wire-reply fields via the ``TeakRewardKey...`` constants below
+ * (e.g. ``userInfo[TeakRewardKeyJwtToken]``). Launch-data attribution fields
+ * use the existing ``teakCamelCase`` keys (e.g. ``teakRewardId``,
+ * ``teakNotifId``, ``teakScheduleId``).
  */
 extern NSString* _Nonnull const TeakOnRewardJwtIssued;
 
@@ -36,6 +41,8 @@ extern NSString* _Nonnull const TeakOnRewardJwtIssued;
  * The notification's userInfo carries the event id (used by the SDK's
  * click-time poll loop) plus the reward and attribution context so the host
  * game can render an optimistic surface while the claim is in flight.
+ *
+ * Access the wire-reply fields via the ``TeakRewardKey...`` constants below.
  */
 extern NSString* _Nonnull const TeakOnRewardClaimPending;
 
@@ -45,16 +52,43 @@ extern NSString* _Nonnull const TeakOnRewardClaimPending;
  * state (completed or failed) for a previously-pending claim.
  *
  * The notification's userInfo carries the polled status, the customer's
- * server response, and the launch-data attribution fields.
+ * server response, and the launch-data attribution fields. Access the
+ * wire-reply fields via the ``TeakRewardKey...`` constants below.
  *
  * Delivery is at-least-once: if the SDK's `/claim_ack` post fails (network
  * blip, app backgrounded mid-flight), the server's session-start sweep will
  * resurface the same resolved claim on the next launch and fire this
- * notification again with the same `event_id`. Host games SHOULD idempotent-
- * key any reward grant on the `event_id` field of the userInfo dictionary
- * so a duplicate delivery doesn't grant the reward twice.
+ * notification again with the same ``TeakRewardKeyEventId``. Host games
+ * SHOULD idempotent-key any reward grant on that field so a duplicate
+ * delivery doesn't grant the reward twice.
  */
 extern NSString* _Nonnull const TeakOnRewardClaimResolved;
+
+/**
+ * Keys for the ``userInfo`` dictionary on ``TeakOnRewardJwtIssued``,
+ * ``TeakOnRewardClaimPending``, and ``TeakOnRewardClaimResolved``.
+ *
+ * The wire reply from the server uses snake_case field names; these
+ * constants are the canonical access path so host games don't depend on
+ * the literal strings. Launch-data attribution fields (``teakRewardId``,
+ * ``teakNotifId``, ``teakScheduleId``, etc.) keep their existing
+ * ``teakCamelCase`` names.
+ *
+ * ``TeakRewardKeyId`` (the snake_case ``teak_reward_id`` from the server)
+ * is the reward the server authoritatively granted on this click; the
+ * launch-data ``teakRewardId`` field is the reward this launch was
+ * *attributed to* (from the URL or notification payload). The two can
+ * differ on proxy reward routes, and both are preserved on the same
+ * userInfo dictionary as distinct fields.
+ */
+extern NSString* _Nonnull const TeakRewardKeyEventId;
+extern NSString* _Nonnull const TeakRewardKeyStatus;
+extern NSString* _Nonnull const TeakRewardKeyJwtToken;
+extern NSString* _Nonnull const TeakRewardKeyReward;
+extern NSString* _Nonnull const TeakRewardKeyClaimSource;
+extern NSString* _Nonnull const TeakRewardKeyId;
+extern NSString* _Nonnull const TeakRewardKeyCustomerResponse;
+extern NSString* _Nonnull const TeakRewardKeyCustomerStatusCode;
 
 /**
  * Use this named notification to listen for when your app receives a Teak notification while in the foreground.
--- a/Teak/Teak.h
+++ b/Teak/Teak.h
@@ -18,6 +18,10 @@ extern NSString* _Nonnull const TeakNotificationAppLaunch;
  */
 extern NSString* _Nonnull const TeakOnReward;
 
+extern NSString* _Nonnull const TeakOnRewardJwtIssued;
+extern NSString* _Nonnull const TeakOnRewardClaimPending;
+extern NSString* _Nonnull const TeakOnRewardClaimResolved;
+
 /**
  * Use this named notification to listen for when your app receives a Teak notification while in the foreground.
  * 	[[NSNotificationCenter defaultCenter] addObserver:self
--- a/Teak/TeakClaimPoll.m
+++ b/Teak/TeakClaimPoll.m
@@ -1,0 +1,29 @@
+#import "TeakClaimPoll.h"
+
+@implementation TeakClaimPoll
+
++ (NSTimeInterval)nextDelayAfter:(NSUInteger)attempt
+                    initialDelay:(NSTimeInterval)initialDelay
+                         ceiling:(NSTimeInterval)ceiling {
+  return 0.0;
+}
+
++ (BOOL)isTerminalStatus:(NSString*)status {
+  return NO;
+}
+
++ (NSDictionary*)buildResolvedUserInfoForReply:(NSDictionary*)reply
+                                withLaunchData:(TeakAttributedLaunchData*)launchData {
+  return nil;
+}
+
++ (void)startPollForEventId:(NSString*)eventId
+                 launchData:(TeakAttributedLaunchData*)launchData
+               initialDelay:(NSTimeInterval)initialDelay
+                    ceiling:(NSTimeInterval)ceiling {
+}
+
++ (void)cancelAllPolls {
+}
+
+@end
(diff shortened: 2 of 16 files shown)


Teak/Teak.h
```diff
@@ -18,6 +18,44 @@ extern NSString* _Nonnull const TeakNotificationAppLaunch;
...
+ */
+extern NSString* _Nonnull const TeakOnRewardClaimPending;
+
+/**
+ * Use this named notification to listen for the resolved-claim response from
+ * the SDK's click-time poll loop. Fires when the server reaches a terminal
+ * state (completed or failed) for a previously-pending claim.
+ *
+ * The notification's userInfo carries the polled status, the customer's
+ * server response, and the launch-data attribution fields.
```
AlexSc: Do we document what the keys are anywhere? Or provide constants for accessing them?

Teak/TeakSession.m
```diff
@@ -817,6 +861,12 @@ + (void)checkLaunchDataForDeepLinkAndDispatchEvents:(nonnull TeakLaunchData*)lau
     } else if (newValue == [TeakSession Expiring]) {
       self.endDate = [[NSDate alloc] init];
 
+      // Click-time claim polls don't outlive their session — they fire from
+      // an in-session click and the resolved-event payload is read from the
+      // launch-data state we still hold. The session-start sweep handles
+      // claims that resolve after the app closes.
+      [TeakClaimPoll cancelAllPolls];
```
AlexSc: A session can transition back from expiring to IdentifiedUser -- e.g. a user opens the notification center briefly.

That shouldn't abort polling unless we can resume it. I'm inclined to move this to expired, or extend this to parking polling and cancel on the move to expired and resume on returning the session to active.

Teak/TeakSession.m
```diff
@@ -656,25 +657,68 @@ - (void)identifyUserInfoHasChanged {
...
+  if (launchData != nil) {
+    [userInfo addEntriesFromDictionary:[launchData to_h]];
+  }
+  [userInfo addEntriesFromDictionary:reply];
+
+  // Default initial delay 2s, ceiling 30s — chosen to keep the first poll close
+  // to claim-completion latency while preventing busy-loops if the customer
+  // backend is slow.
+  static const NSTimeInterval kClaimPollInitialDelay = 2.0;
+  static const NSTimeInterval kClaimPollCeiling = 30.0;
```
AlexSc: I want these to come from server config. We can use these values as fallbacks.

ProductOps: Please create a tracking issue to surface these from server config. They can be hardcoded, I just want to lock in the contract.

Teak/TeakSession.m
```diff
@@ -656,25 +657,68 @@ - (void)identifyUserInfoHasChanged {
...
+                            initialDelay:kClaimPollInitialDelay
+                                 ceiling:kClaimPollCeiling];
+    }
+  }
+
+  NSNotification* note = [NSNotification notificationWithName:notificationName
+                                                       object:nil
+                                                     userInfo:userInfo];
+  [TeakSession whenUserIdIsReadyRun:^(TeakSession* session) {
+    [[NSNotificationCenter defaultCenter] postNotificationName:notificationName
```
AlexSc: Having separate paths for production and testing concerns me. I'd prefer to consolidate testing on to what we do in production. I can be talked into allowing the split if we can construct what we're passing to the notification center out of the previously built note.

Teak/TeakSession.m
```diff
@@ -656,25 +657,68 @@ - (void)identifyUserInfoHasChanged {
...
+  }
+  [userInfo addEntriesFromDictionary:reply];
+
+  // Default initial delay 2s, ceiling 30s — chosen to keep the first poll close
+  // to claim-completion latency while preventing busy-loops if the customer
+  // backend is slow.
+  static const NSTimeInterval kClaimPollInitialDelay = 2.0;
+  static const NSTimeInterval kClaimPollCeiling = 30.0;
+
+  if ([wireStatus isEqualToString:@"claim_pending"]) {
```
AlexSc: Not a fan of checking wireStatus twice for the same string

Teak/TeakClaimPoll.m
```diff
@@ -0,0 +1,260 @@
...
+
+  NSMutableURLRequest* request = [NSMutableURLRequest requestWithURL:components.URL];
+  request.HTTPMethod = @"POST";
+  request.HTTPBody = [NSJSONSerialization dataWithJSONObject:payload options:0 error:nil];
+  [request setValue:@"application/json" forHTTPHeaderField:@"Content-Type"];
+
+  TeakLog_i(@"claim_ack.request.send", @{@"event_id" : eventId});
+
+  NSURLSession* urlSession = [Teak URLSessionWithoutDelegate];
+  NSURLSessionDataTask* task = [urlSession dataTaskWithRequest:request
```
AlexSc, line 337: I would like to make this retriable. While we do ask our integrators to make their resolution handler idempotent we shouldn't force that upon them until strictly necessary. An error we can retry and save them the hassle. This goes double for something they have no ability to retry themselves.

Teak/TeakClaimPoll.m
```diff
@@ -0,0 +1,260 @@
...
+  // path).
+  [[TeakClaimPoll activePolls] setObject:kInFlightSentinel forKey:eventId];
+
+  // Snapshot the generation for the completion's session-still-alive check.
+  uint64_t generation = sPollGeneration;
+
+  [TeakClaimPoll sendClaimStatusRequestForEventId:eventId
+                                       completion:^(NSDictionary* reply) {
+                                         dispatch_async(dispatch_get_main_queue(), ^{
+                                           if (generation != sPollGeneration) {
```
AlexSc: Instead of generations, I think we should keep a weak reference to the session that started us and call it stale if the session doesn't match when we go to fireResolved.

This is kind of a continuation of noting that a session can go from Expiring back to active -- that's a case we need to handle, and right now we wouldn't.

Teak/TeakClaimPoll.m
```diff
@@ -0,0 +1,260 @@
...
+                                             // Session expired between request-send and reply.
+                                             // Drop the reply; cancelAllPolls already cleared the
+                                             // dictionary entry.
+                                             TeakLog_i(@"claim_poll.reply.stale", @{@"event_id" : eventId});
+                                             return;
+                                           }
+
+                                           NSString* status = reply[@"status"];
+                                           if ([TeakClaimPoll isTerminalStatus:status]) {
+                                             [[TeakClaimPoll activePolls] removeObjectForKey:eventId];
```
AlexSc: Let's not remove from active polls until acked. Less "Active polls" and more "yea, I'm on it"

Teak/TeakClaimPoll.m
```diff
@@ -0,0 +1,260 @@
...
+  NSTimeInterval ceiling = [userInfo[@"ceiling"] doubleValue];
+  id launchDataValue = userInfo[@"launch_data"];
+  TeakAttributedLaunchData* launchData = (launchDataValue == [NSNull null]) ? nil : launchDataValue;
+
+  // Replace the timer slot with the in-flight sentinel — keeps dedupe truthy
+  // for the duration of the request so a re-entrant +startPollForEventId:
+  // sees an existing poll. The slot is cleared by the completion handler
+  // (terminal → fire+ack path) or replaced by the next timer (continue-poll
+  // path).
+  [[TeakClaimPoll activePolls] setObject:kInFlightSentinel forKey:eventId];
```
AlexSc: Is there any reason to use a sentinel instead of just leaving the timer object in place? Feels like extra ceremony and I'm not sure what it's for. Is there any problem with invalidating a fired timer, or checking if the timer already fired?

Teak/TeakClaimPoll.m
```diff
@@ -0,0 +1,260 @@
...
+                                                                         initialDelay:initialDelay
+                                                                              ceiling:ceiling];
+                                         });
+                                       }];
+}
+
++ (void)fireResolvedAndAck:(NSDictionary*)reply
+                   eventId:(NSString*)eventId
+                launchData:(TeakAttributedLaunchData*)launchData {
+  NSDictionary* userInfo = [TeakClaimPoll buildResolvedUserInfoForReply:reply
```
AlexSc, line 197: I'd like to log here, info level, just to be sure that we made it.

AlexSc (changes requested):
I didn't review the tests, since I figure there may be some rework.

High level -- on the right track, but needs a simplification pass and there are some misses from not fully understanding the session lifecycle. Any session lifecycle state transition that we permit can and will happen, and we need to handle those cases.

Teak/TeakSession.m
TeakBuilds: [productops] Filed as C-721 (https://linear.app/teakio/issue/C-721) — "Surface JWT-mode click-poll backoff values from server config." Hardcoded values stay in PR #135 as the documented fallback contract; the server-config field is the follow-on. Cross-references the SDK B click-poll convention so Android (C-701) and JS (C-706) inherit the same fallback shape.

Teak/TeakClaimPoll.m
```diff
@@ -0,0 +1,346 @@
...
+    return;
+  }
+
+  TeakSession* originatingSession = [TeakSession currentSessionOrNil];
+
+  dispatch_async(dispatch_get_main_queue(), ^{
+    NSMutableDictionary* claims = [TeakClaimPoll inflightClaims];
+    if (claims[eventId] != nil) {
+      // SDK is already responsible for delivering this event_id. Re-entrant
+      // start is a no-op — the existing claim continues with its own session
```
AlexSc, line 93: I'd like an info log for this case

Teak/TeakClaimPoll.m
```diff
@@ -0,0 +1,346 @@
...
+                                                       userInfo:@{@"event_id" : claim.eventId}
+                                                        repeats:NO];
+}
+
++ (void)pollTimerFired:(NSTimer*)timer {
+  NSString* eventId = timer.userInfo[@"event_id"];
+  TeakInflightClaim* claim = [TeakClaimPoll inflightClaims][eventId];
+  if (claim == nil) {
+    // Cancelled between schedule and fire. Nothing to do.
+    return;
```
AlexSc, line 147: Info log

Teak/TeakClaimPoll.m
```diff
@@ -0,0 +1,346 @@
...
+                                         [TeakClaimPoll handleClaimStatusReply:reply forEventId:eventId];
+                                       });
+                                     }];
+}
+
++ (void)handleClaimStatusReply:(NSDictionary*)reply forEventId:(NSString*)eventId {
+  TeakInflightClaim* claim = [TeakClaimPoll inflightClaims][eventId];
+  if (claim == nil) {
+    // Cancelled while the request was in flight. Drop the reply silently.
+    return;
```
AlexSc, line 168: Info log

Teak/TeakClaimPoll.m
```diff
@@ -0,0 +1,346 @@
...
+    dispatch_async(dispatch_get_main_queue(), ^{
+      [TeakClaimPoll attemptAckForEventId:claim.eventId session:session];
+    });
+  }];
+}
+
++ (void)attemptAckForEventId:(NSString*)eventId session:(TeakSession*)session {
+  TeakInflightClaim* claim = [TeakClaimPoll inflightClaims][eventId];
+  if (claim == nil) {
+    // Cancelled between resolve and ack. Drop.
```
AlexSc, line 218: Info log

Teak/TeakSession.m
```diff
@@ -656,25 +657,67 @@ - (void)identifyUserInfoHasChanged {
...
+/// status. Returns the notification it dispatched (for testability —
+/// production callers ignore the return value; the same notification is
+/// posted once the user id is ready).
++ (NSNotification*)dispatchClickResponse:(NSDictionary*)reply forLaunchData:(TeakAttributedLaunchData*)launchData {
+  // Default initial delay 2s, ceiling 30s — chosen to keep the first poll close
+  // to claim-completion latency while preventing busy-loops if the customer
+  // backend is slow. Tracked for promotion to remote config in C-721; until
+  // then these are the cross-SDK fallback values.
+  static const NSTimeInterval kClaimPollInitialDelay = 2.0;
+  static const NSTimeInterval kClaimPollCeiling = 30.0;
```
AlexSc: Let's pull server values in on this pr. Coordinate with ProductOps on the field naming as C-721 goes through dev

AlexSc: But we keep these as a backup if there are no server values. They should come in through settings

AlexSc (changes requested):

Teak/TeakClaimPoll.m
AlexSc: Whoops, missed this one

Teak/TeakSession.m
```diff
@@ -656,25 +657,71 @@ - (void)identifyUserInfoHasChanged {
...
+/// status. Returns the notification it dispatched (for testability —
+/// production callers ignore the return value; the same notification is
+/// posted once the user id is ready).
++ (NSNotification*)dispatchClickResponse:(NSDictionary*)reply forLaunchData:(TeakAttributedLaunchData*)launchData {
+  // Cross-SDK fallback for the poll cadence. Used only when there's no
+  // current session (test paths) or no remote configuration on it. Live
+  // sessions read the values from TeakRemoteConfiguration, which itself
+  // initializes to the same defaults and overrides them from the
+  // /games/.../settings.json reply.
+  static const NSTimeInterval kClaimPollInitialDelayFallback = 2.0;
```
AlexSc: Can we centralize these in remote config instead of redeclaring?

Teak/TeakClaimPoll.m
TeakBuilds: [worker-698] The `claim_resolved.delivered` info log is already present at `TeakClaimPoll.m:199`, inside the `whenUserIdIsReadyRun:` block right after the `postNotificationName:` for `TeakOnRewardClaimResolved`. Added in round 2 (commit `19ee453`) per the original finding #10. If you meant a different spot — e.g. an entry-point log at the top of `+fireResolvedAndAckForClaim:reply:` so the trace shows "received terminal status, queued for delivery" separately from "actually delivered" — happy to add it; just confirm.

AlexSc: Yea, I would like at the entry just in case we don't get to the user id ready

AlexSc (approved):
Couple of minor bits I'd like cleaned up before merge, but I don't need to rereview from those

PR #136, opened 2026-04-28
C-699: iOS session-start /claims sweep

TeakBuilds:
Closes https://linear.app/teakio/issue/C-699/ios-sdk-session-start-claims-sweep

## Summary

After user-identify, the iOS SDK pulls `GET /claims?teak_app_id=…&clicking_user_id=…` and dispatches each unacked claim through the same in-flight tracker the click-time path (PR #135) already uses. Terminal entries fire `TeakOnRewardClaimResolved` and run the existing retriable ack ceremony; pending entries enroll into the click-time poll loop without re-firing `TeakOnRewardClaimPending` (point-in-time event, not history-replay). Idempotency is the in-flight dictionary's `event_id` key — sweep entries for ids already mid-poll, mid-request, or post-resolve awaiting ack are no-ops.

## Key decisions

- **Refactored `TeakInflightClaim` to store the eleven-key attribution dict** instead of a `TeakAttributedLaunchData`. Click-time flattens its launch-data once at start, so both surfaces share the same fire/ack machinery. Public launch-data-taking helpers stay as thin wrappers — no API churn for callers.
- **Tolerant `session_attribution` unpack**: accept either an inline dict or a JSON string (matches the click-POST mint shape). Robust to either taro encoding choice.
- **Sweep enrolls terminal entries into the in-flight dictionary** before firing resolved+ack, so the existing bounded retry budget and originating-session pin apply uniformly to both click-time and sweep entries.
- **Entry point**: `TeakSession`'s `UserIdentified` KVO branch, after `processAttributionAndDispatchEvents`, gated on `oldValue != Expiring`. Sweep runs on fresh user-identify passes only — the Expiring → UserIdentified flicker is a same-session resume and re-running the sweep there would double-fire resolved events for claims whose ack-2xx hasn't yet committed server-side. The at-least-once contract on `TeakOnRewardClaimResolved` is anchored to fresh launches.
- **Strip on resolved-event userInfo**: `session_attribution` (already unpacked into top-level attribution keys), `created_at`, and `completed_at` are stripped from the wire reply before merging. They aren't part of the documented public surface; the strip set is a static `NSSet` for one-line additions if more bookkeeping fields land later.
- **C-721 inheritance**: the sweep's pending-poll cadence reads `claimPollInitialDelay` / `claimPollCeiling` from `TeakRemoteConfiguration`, same as the click-time path.

## Test plan

- Spec-first commit (red) precedes implementation (green); reviewer can read both diffs in sequence.
- New `SessionStartSweepTests` covers nine invariants on the dispatcher: dict-attribution merge helper (with the two reward-id flavors coexisting), terminal/pending in-flight registration, click-time dedupe, no `TeakOnRewardClaimPending` re-fire on sweep, empty-list and malformed-entry tolerance, string-encoded `session_attribution` unpack.
- New `JwtModeRewardEventsTests` invariant: `testResolvedEventStripsServerBookkeepingAndRawAttributionBlob`.
- Full local run: 172 tests in AutomatedTests target, all green; 1 UI test, green.

## Readers left behind

- End-to-end NSNotificationCenter delivery on the sweep's terminal-claim fire is exercised at integration, not unit. The fire path goes through `whenUserIdIsReadyRun:`, which queues until the test session reaches UserIdentified — not driven in the test environment. Helper-level coverage (`testBuildResolvedUserInfoMergesAttributionDictAndReply`, `testResolvedEventStripsServerBookkeepingAndRawAttributionBlob`) plus the in-flight registration tests cover the merge correctness and enrollment; the actual NSNotificationCenter post is integration-level.
- Ack POST uses the `whenUserIdIsReadyRun:` current session, not `claim.originatingSession`. Pre-existing defect (TeakClaimPoll.m, shipped in PR #135) folded into C-722 per ProductOps — out of scope for this PR.

### DIFF
--- a/Automated/AutomatedTests/SessionStartSweepTests.m
+++ b/Automated/AutomatedTests/SessionStartSweepTests.m
@@ -1,0 +1,287 @@
+#import <XCTest/XCTest.h>
+
+#import <Teak/Teak.h>
+
+#import "TeakClaimPoll.h"
+#import "TeakLaunchData.h"
+
+@import OCHamcrest;
+@import OCMockito;
+
+@interface TeakConfiguration : NSObject
++ (BOOL)configureForAppId:(NSString*)appId andSecret:(NSString*)appSecret;
+@end
+
+// Internal entry points the sweep dispatcher exposes for unit-level coverage.
+// `dispatchSweptClaims:session:` is the wire-free seam — production callers
+// reach it through `+startSweep`, which composes the GET /claims request and
+// hands the parsed array off to this dispatcher.
+@interface TeakClaimPoll (Testing)
++ (NSMutableDictionary*)inflightClaims;
++ (NSDictionary*)buildResolvedUserInfoForReply:(NSDictionary*)reply
+                                withAttribution:(NSDictionary*)attribution;
++ (void)dispatchSweptClaims:(NSArray*)claims session:(id)session;
+@end
+
+@interface SessionStartSweepTests : XCTestCase
+@end
+
+@implementation SessionStartSweepTests
+
++ (void)setUp {
+  [super setUp];
+  @try {
+    [TeakConfiguration configureForAppId:@"test-app" andSecret:@"test-secret"];
+  } @catch (NSException* e) {
+    // Already initialized — fine.
+  }
+}
+
+- (void)setUp {
+  [super setUp];
+  [TeakClaimPoll cancelAllPolls];
+  [self drainMainQueue];
+}
+
+- (void)tearDown {
+  [TeakClaimPoll cancelAllPolls];
+  [self drainMainQueue];
+  [super tearDown];
+}
+
+- (void)drainMainQueue {
+  XCTestExpectation* e = [self expectationWithDescription:@"main-queue drain"];
+  dispatch_async(dispatch_get_main_queue(), ^{
+    [e fulfill];
+  });
+  [self waitForExpectations:@[ e ] timeout:1.0];
+}
+
+#pragma mark - Resolved-userInfo build (dict-taking helper)
+
+/// The dict-taking helper unpacks the eleven-key session_attribution blob and
+/// merges it with the wire reply at fire time. Reply wins on key collision —
+/// same merge order as the launchData-taking helper used by the click-time
+/// path. The two reward-id flavors (teakRewardId attribution vs. teak_reward_id
+/// authoritative) coexist on the same userInfo as distinct fields.
+- (void)testBuildResolvedUserInfoMergesAttributionDictAndReply {
+  NSDictionary* attribution = @{
+    @"launch_link" : @"teaktest-app://chest",
+    @"teakNotifId" : @"2048153148060669486",
+    @"teakScheduleId" : @"2046986133304291328",
+    @"teakScheduleName" : @"daily_promo_2026q2",
+    @"teakCreativeId" : @"2046986561123301779",
+    @"teakCreativeName" : @"summer_sale_v3",
+    @"teakRewardId" : @"2048153148060669138",
+    @"teakChannelName" : @"ios_push",
+    @"teakDeepLink" : [NSNull null],
+    @"teakOptOutCategory" : @"teak",
+    @"teakSystemActivityId" : [NSNull null],
+  };
+  NSDictionary* reply = @{
+    @"event_id" : @"evt-resurfaced-1",
+    @"status" : @"completed",
+    @"reward" : @{@"gems" : @25},
+    @"customer_response" : @"{\"ok\":true}",
+    @"customer_status_code" : @200,
+    @"teak_reward_id" : @"2048153148060669999",
+  };
+
+  NSDictionary* userInfo = [TeakClaimPoll buildResolvedUserInfoForReply:reply
+                                                          withAttribution:attribution];
+  XCTAssertNotNil(userInfo);
+  XCTAssertEqualObjects(userInfo[@"event_id"], @"evt-resurfaced-1");
+  XCTAssertEqualObjects(userInfo[@"status"], @"completed");
+  XCTAssertEqualObjects(userInfo[@"reward"], @{@"gems" : @25});
+  XCTAssertEqualObjects(userInfo[@"customer_status_code"], @200);
+
+  XCTAssertEqualObjects(userInfo[@"teakNotifId"], @"2048153148060669486");
+  XCTAssertEqualObjects(userInfo[@"teakScheduleId"], @"2046986133304291328");
+  XCTAssertEqualObjects(userInfo[@"teakCreativeName"], @"summer_sale_v3");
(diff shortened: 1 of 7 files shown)


AlexSc:
ProductOps: Please kick off C-723

AlexSc:
Walk me through how this all works on a start that's claiming a reward. What's the lifecycle, what happens when, what's the risk of races? Keep the reply concise please.

Teak/TeakClaimPoll.m
```diff
@@ -63,22 +72,63 @@ + (BOOL)isTerminalStatus:(NSString*)status {
...
+  return [TeakClaimPoll buildResolvedUserInfoForReply:reply
+                                       withAttribution:(launchData != nil ? [launchData to_h] : nil)];
+}
+
++ (NSDictionary*)buildResolvedUserInfoForReply:(NSDictionary*)reply
+                               withAttribution:(NSDictionary*)attribution {
   NSMutableDictionary* userInfo = [[NSMutableDictionary alloc] init];
-  if (launchData != nil) {
-    [userInfo addEntriesFromDictionary:[launchData to_h]];
+  if ([attribution isKindOfClass:[NSDictionary class]]) {
```
AlexSc: Could attribution be something other than a dictionary? Check feels weird.

Teak/TeakClaimPoll.m
```diff
@@ -63,22 +72,63 @@ + (BOOL)isTerminalStatus:(NSString*)status {
...
 #pragma mark - Lifecycle
 
 + (void)startPollForEventId:(NSString*)eventId
                  launchData:(TeakAttributedLaunchData*)launchData
                initialDelay:(NSTimeInterval)initialDelay
                     ceiling:(NSTimeInterval)ceiling {
+  // Click-time path: flatten the launch-data to its eleven-key wire shape
+  // once at start so all subsequent reads route through the same dict the
+  // session-start sweep uses.
+  NSDictionary* attribution = launchData != nil ? [launchData to_h] : nil;
```
AlexSc: Wouldn't [nil to_h] just be nil anyways?

Automated/AutomatedTests/SessionStartSweepTests.m
```diff
@@ -0,0 +1,288 @@
...
+  [self waitForExpectations:@[ e ] timeout:1.0];
+}
+
+#pragma mark - Resolved-userInfo build (dict-taking helper)
+
+/// The dict-taking helper unpacks the eleven-key session_attribution blob and
+/// merges it with the wire reply at fire time. Reply wins on key collision —
+/// same merge order as the launchData-taking helper used by the click-time
+/// path. The two reward-id flavors (teakRewardId attribution vs.
+/// teak_reward_id authoritative) coexist on the same userInfo as distinct
```
AlexSc: How do we end up with different attribution and authoritative ids?

AlexSc (changes requested):

Teak/TeakClaimPoll.m
TeakBuilds: [worker-699] Fixed in d6ddb8c — dropped the kind-check, just nil-checking now. Same simplification applied to the `reply` arm for symmetry.

Teak/TeakClaimPoll.m
TeakBuilds: [worker-699] Fixed in d6ddb8c — calling `[launchData to_h]` directly. The dict-taking helper already nil-checks the result.

Automated/AutomatedTests/SessionStartSweepTests.m
TeakBuilds: [worker-699] The framing is from `Teak.h:79-82` (shipped in PR #135) — proxy-reward routes resolve a click on URL-attributed reward A to an authoritatively-granted reward B at server grant time (e.g. A is exhausted and substitutes to B, or A is an A/B variant family that picks one). The fixture uses distinct ids to prove the merge keeps both keys without aliasing. Happy to align if proxy-reward routes are stale / not in scope — easiest fix is to make the fixture ids equal and tighten the comment to "exercises the merge contract" without claiming a real-world scenario.

AlexSc: Ugh, okay. The naming is confusing. Thoughts?

TeakBuilds: [worker-699] Agreed it is. The discriminator is case-only (`teakRewardId` vs `teak_reward_id`), which is genuinely the worst kind of typo-trap. Three honest paths, none free:

1. **Status quo, tighter docs.** Keep both keys, beef up `Teak.h:79-82` with a one-liner example contrasting attributed vs. granted. Cheapest, but the case-only discriminator stays.

2. **Drop the launch-data flavor from `TeakOnRewardClaimResolved` only.** Reasoning: this event is *about* the resolved server grant; the attributed-reward flavor is already exposed on `TeakNotificationAppLaunch` / `TeakLaunchedFromLink` for the same launch. Host games that need both correlate by `event_id`. Surgical — only the resolved-event userInfo loses `teakRewardId`. Touches Teak.h docs and the click-time path equally; needs Android/JS coordination since this is cross-SDK contract.

3. **Rename the launch-data flavor cross-cuttingly.** `teakRewardId` → `teakAttributedRewardId` (or similar) on every event that carries launch-data attribution. Reads cleanly, eliminates the case-only discriminator. Public surface change — every host game listening on `TeakNotificationAppLaunch` reads `teakRewardId` today. Big blast radius.

My pick if forced: (2). The proxy-reward distinction stays meaningful, but the resolved event becomes single-flavored — `teak_reward_id` only — which is what host games actually need to grant the reward. The attribution still exists, just not redundantly on this event. Want me to spin a proposal commit on this branch (or punt it to a separate PR if the cross-SDK coordination is heavy)?

If keeping both flavors is the call, I will tighten the test comment + the in-file framing to lead with the merge contract, with proxy-reward routes as the why-it-matters footnote rather than the headline.

AlexSc: What's actually happening here is we have a reward id we got in attribution, but then we're granting a different reward id. But, in our existing legacy claim, we would never surface that, right? We'd just show the attributed reward id.

So my gut feeling is we should match the legacy claim semantics -- we only surface the attribution id. Thoughts?

TeakBuilds: [worker-699] Agreed — matched legacy `TeakOnReward` semantics in 1c98ba2. Stripped `teak_reward_id` from the resolved-event wire reply; userInfo carries only `teakRewardId` from launch-data. Tests updated, Teak.h Keys block doc updated to scope `TeakRewardKeyId` to JwtIssued/ClaimPending only. The case-only-discriminator goes away on the resolved event. 172 tests green.

AlexSc: ProductOps: This is the way. Please make sure other SDKs land this shape

AlexSc (approved):

PR #138, opened 2026-05-01
feat(C-683): auto-add TeakLog events as Sentry breadcrumbs

TeakBuilds:
Closes https://linear.app/teakio/issue/C-683/ios-sdk-auto-add-teaklog-events-as-sentry-breadcrumbs

## Summary
- Sentry reports were missing most SDK context because TeakLog_i/TeakLog_e events never appeared as breadcrumbs — only a handful of manually-placed `teak_log_breadcrumb` macro calls did.
- `logEvent:level:eventData:` now adds every event to a session-scoped global breadcrumb store on `TeakRavenLocationHelper`. Works whether or not a `teak_try` scope is active.
- `reportWithHelper:` snapshots the store at report time (via the same serial queue) and includes it in the Sentry payload. Breadcrumbs are not cleared after send — the next exception sees full history.
- Store is bounded at 100 entries (FIFO) to match Sentry's default and keep memory bounded.

## Key decisions
- **Global store, not per-helper:** breadcrumbs accumulate across the entire session so Sentry sees history leading up to a crash, not just events inside one `teak_try` block.
- **Serial dispatch queue (`io.teak.breadcrumbs`):** all array mutations and snapshot reads are serialized on a dedicated queue; dict is built on the caller's thread so the timestamp is captured at call time.
- **`eventType` as Sentry `category`** (e.g. `configuration.device`, `request.send`) — gives useful Sentry groupings. `log_level` is preserved in the data dict.

## Test plan
Three unit tests in `LogTests.m`: breadcrumb lands in the global store with correct category/data; no crash when called outside a `teak_try` scope; cap holds at 100 after 110 events with FIFO eviction verified. All pass: `bundle exec fastlane test`.

## Follow-up scope
- Existing `teak_log_breadcrumb` macro call sites (TeakLink.m, TeakRequest.m, SKPaymentObserver.m) are now redundant — they still write to the global store via the instance method, but the same events go through TeakLog anyway. Flagged for follow-up cleanup.

### DIFF
--- a/Automated/AutomatedTests/LogTests.m
+++ b/Automated/AutomatedTests/LogTests.m
@@ -19,6 +19,10 @@
 @implementation LogTests
 
 - (void)setUp {
+  while ([TeakRavenLocationHelper peekHelper] != nil) {
+    [TeakRavenLocationHelper popHelper];
+  }
+
   self.teakMock = mock([Teak class]);
   [given([self.teakMock enableRemoteLogging]) willReturn:@NO];
   [given([self.teakMock enableDebugOutput]) willReturn:@NO];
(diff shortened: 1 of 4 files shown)


AlexSc:
We _WANT_ breadcrumbs from outside teak_try. We're looking to have session history when something fires.

AlexSc (approved):
LGTM

PR #140, opened 2026-06-15
fix(C-615): capture session under lock in deferred whenUserIdIsReady callbacks

TeakBuilds:
## C-615 — investigate iOS crash

Linear: https://linear.app/teakio/issue/c-615

### Symptom
Kitchen Scramble (com.disney.kitchenscramble 11.2.26/229) reports ~4–6 iOS crashes/day. The Crashlytics report:

```
Crashed: io.teak.sdk.eventProcessingQueue
0  libobjc.A.dylib   objc_retain_x0
1  libobjc.A.dylib   objc_retain
2  +[TeakSession whenUserIdIsReadyRun:]
3  -[TeakCore handleEvent:]
4  __23+[TeakEvent postEvent:]_block_invoke
```

EXC_BAD_ACCESS in `objc_retain` — retaining a freed object reached through `whenUserIdIsReadyRun:`.

### Root cause
`currentSession` is a `__strong` **file-global**. The two deferred callbacks dispatched `^{ block(currentSession); }`. Blocks don't capture globals by value — they reference the global directly, so the retain happened when the block **executed**: on a global concurrent queue, *after* `@synchronized(currentSessionMutex)` had been released.

That unsynchronized `objc_retain` of the strong global races with `currentSession = newSession` (which `objc_release`s the old value) on session reassignment/expiry — a textbook ARC strong-variable data race. It over-releases the session, leaving the global dangling; the next synchronous read in `whenUserIdIsReadyRun:` then crashes in `objc_retain`. The narrow window matches the low, steady crash rate.

### The audit (this is the proof — there's no deterministic repro)
Line numbers below are on this branch's HEAD (after the patch comments).

**Writes to the global — all 3 under `@synchronized(currentSessionMutex)`:**
- `TeakSession.m:533` `logoutReusingCurrentSession:`
- `TeakSession.m:607` `didLaunchWithData:`
- `TeakSession.m:629` `+currentSession`

No setter, no `&currentSession`, no writer outside the file, none using `@synchronized(self)` or no lock. Because every write is under the mutex, copying the global into a strong local under that same lock genuinely closes the race — it doesn't just relocate it.

**Reads from a deferred/async context — only 2, both patched here:**
- `whenUserIdIsReadyRun:` (the async `block(currentSession)` was at :109, now `block(session)` at :113)
- `whenUserIdIsOrWasReadyRun:` (was :123, now :131)

The third async block (`whenDeviceIsAwakeRun:`) calls `block()` and reads no global; the `setState:` drain at :797 passes `block(self)`; every other read of `currentSession` is synchronous and already inside `@synchronized(currentSessionMutex)`.

Confirmed present on `4.3-stable` HEAD (3b0b07a) — a real code change, not regression-only.

### The fix
Bind `TeakSession* session = currentSession;` inside the lock and have the block capture the local. Two sites. Also fixes a latent correctness bug: the callback now runs with the session that was current when it was queued, not whatever is current when the async fires.

Heavier options (serial-queue rewrite, atomic accessor) were ruled out — the mutex already covers every write and synchronous read, so only the two async re-reads are the hole; widening the locking would tangle with the existing nested `@synchronized(self)` / KVO-on-`setState:` and is unjustified risk for a point release.

### Test
`TeakSessionDeferredCallbackTests` drives the `currentSession` global (external linkage) with OCMockito stand-ins, enqueues a callback, swaps the global, and asserts the callback received the session current **at enqueue**. Three cases: `whenUserIdIsReadyRun:`, `whenUserIdIsOrWasReadyRun:`, and the OrWas-specific `Expiring && previousState == UserIdentified` gate. The two `UserIdentified` cases were verified to fail 512/512 on the old code; all three pass on the fix. It documents the capture semantics rather than reproducing the data race itself — a true crash repro needs ThreadSanitizer, which this project's schemes don't enable.

### Follow-up (out of scope, flagged for the lead)
`currentState` is `nonatomic`, written under `@synchronized(self)` in `setState:` but read under `currentSessionMutex` elsewhere — inconsistent locking. Benign today (state values are singletons), but worth a separate issue.

### DIFF
--- a/Automated/Automated.xcodeproj/project.pbxproj
+++ b/Automated/Automated.xcodeproj/project.pbxproj
@@ -31,6 +31,7 @@
 		43B4618F22A1EB54004AC636 /* AdSupport.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 43B4618E22A1EB54004AC636 /* AdSupport.framework */; };
 		43B4619922A20FDC004AC636 /* AdSupport.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 43B4618E22A1EB54004AC636 /* AdSupport.framework */; };
 		43B4619D22A20FE4004AC636 /* StoreKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 43B4617F22A1EB05004AC636 /* StoreKit.framework */; };
+		440EB1A255029FAF5B1ACEB9 /* TeakSessionDeferredCallbackTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 2286E8051950D4EC6E1C79EF /* TeakSessionDeferredCallbackTests.m */; };
 		55AAEB0EF3B9E90616642C8E /* LiveActivityUpdateSchedulingTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 741732706353CEBB41D0AE37 /* LiveActivityUpdateSchedulingTests.m */; };
 		67ED924AB1318D0336E94BE6 /* TeakAttributedLaunchDataEnrichmentTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 864427BAEA58D0DAC8B50E43 /* TeakAttributedLaunchDataEnrichmentTests.m */; };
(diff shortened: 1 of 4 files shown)


AlexSc (approved):
Nit, keep comments timeless. Approval means fix nit then you're good to merge

PR #143, opened 2026-06-16
Emit observable "exception" log event on caught exception (C-843)

TeakBuilds:
## Summary
- On a caught SDK exception (`teak_catch_report` → `reportWithHelper`), iOS now emits an observable `"exception"` TeakLog event, so exceptions are visible to game code (Unity `OnLogEvent`) and remote TeakLog consumers — parity with Android's `log.exception()`.
- `event_data` matches Android's `Raven.throwableToMap`: `type` ← `NSException.name`, `value` ← `NSException.reason`.

## Key decisions
- Emit from `reportWithHelper` (+ `reportUncaughtException`) using a fresh dict, so the Sentry report payload is byte-for-byte unchanged.
- Dropped `send()`'s old `TeakLog_e(@"exception", self.payload)` — it shipped the raw Sentry blob (no top-level `type`/`value`, so C-739 couldn't match) and would otherwise double-emit. Confirmed no backend consumer depends on the old shape.
- `{type, value}` only — `module` has no faithful iOS analog and the iOS stacktrace frame shape diverges from Android's (and the full stacktrace already goes to Sentry).

## Test plan
- `bundle exec fastlane test` — green.
- `testReportTestExceptionEmitsObservableExceptionEvent` drives the real `reportTestException` path through a real `Teak` + capturing `logListener` (no network) and asserts `event_type == "exception"`, `type == "ReportTestException"`, `value == "Version: 4.3.13-test"`.
- `testExceptionLogEventDataMapsNameToTypeAndReasonToValue` / `…IsNilSafeForMissingReason` cover the shaping helper, including a nil reason.

Linear: https://linear.app/teakio/issue/C-843/teak-ios-emit-an-observable-log-event-on-caught-exception-parity-with

### DIFF
--- a/Automated/AutomatedTests/TeakRavenTests.m
+++ b/Automated/AutomatedTests/TeakRavenTests.m
@@ -14,7 +14,14 @@
 @interface Teak ()
 @property (strong, nonatomic) TeakConfiguration* _Nonnull configuration;
 @property (strong, nonatomic) TeakLog* _Nonnull log;
+@property (strong, nonatomic) TeakRaven* _Nonnull sdkRaven;
+@property (strong, nonatomic, readwrite) NSString* _Nonnull sdkVersion;
+- (void)reportTestException;
 @end
+
+// The shared instance is a plain global with external linkage, so a test can
+// install a real Teak to exercise the singleton-routed exception path.
+extern Teak* _Nullable _teakSharedInstance;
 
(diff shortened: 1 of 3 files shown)


AlexSc (approved):
Let's create a followup issue (for next patch release) to log on uncaught exceptions on iOS and Android

PR #144, opened 2026-06-16
Add public processDeferredDeepLink: helper (C-744)

TeakBuilds:
## What

Adds a public `-[Teak processDeferredDeepLink:]` for the deferred-deep-link flow (C-744): an upstream attribution SDK (e.g. Singular) hands the app an unresolved Teak universal link *after* launch, and Teak resolves it as if the user had tapped it — session start, reward claim, and deep link routing. `TeakOnReward` / `TeakLaunchedFromLink` observers and registered `TeakLink` routes all fire as normal.

Swift: `Teak.sharedInstance()?.processDeferredDeepLink(url)` — no bridging-header forward decl, no `continue:` rename, no `NSUserActivity` knowledge needed. This obsoletes the stopgap workaround documented on C-744.

## How

Thin wrapper over the existing `application:continueUserActivity:restorationHandler:` path — it manufactures the same `NSUserActivity` the OS hands us for a universal-link tap (`NSUserActivityTypeBrowsingWeb` + `webpageURL`) and feeds it through. No routing logic is duplicated (`TeakSceneHooks.m:82` manufactures it the same way).

Design notes:
- **void return** — the underlying call returns `YES` unconditionally, so there's no meaningful "handled?" signal.
- **nil `application:` arg** — the receiver ignores it (`TeakUnused`), and nil avoids touching `[UIApplication sharedApplication]` off the main thread. Attribution-SDK callbacks may fire off-main, and the underlying path is already thread-safe (`TeakSession` mutex + background op-queue), so the helper is safe from any thread. A code comment warns against "fixing" this back to `sharedApplication`.
- **nil `url`** is a no-op.

## Commits

1. Alex's C-744 investigation WIP (Sample DDL buttons, `teak-dev2` domain, reward-status alert) + the `TeakReward.h`/`TeakSceneHooks.h` public-header build fix.
2. The public helper (`Teak.h` + `Teak.m`) + `4.3.13` changelog `new` entry + Sample switched to the helper (openURL button relabeled as the documented negative).

## Verification

- **Sample app builds** for simulator — the call resolves against `<Teak/Teak.h>` alone, confirming the helper is public.
- **`fastlane test`: 129 tests, 0 failures** — no regression.
- The underlying `continueUserActivity` path was e2e-verified in Alex's C-744 investigation (session start + `/slots/ddl_test` route + `status = already_clicked` reward on the test link). On-device re-validation of the wrapper itself defers to ship-time QA.

Targets `4.3-stable` (ships in 4.3.13).

Closes C-744.

### DIFF
--- a/Sample/Sample/ViewController.m
+++ b/Sample/Sample/ViewController.m
@@ -99,23 +99,23 @@ extern void TeakAddOperationToQueue(NSOperation* op);
   }
 }
 
-// Singular hands an unresolved Teak universal link to the app post-launch and we
-// need it to start a session, claim the reward, and route the deep link. C-744
-// test path A: hand the URL to UIApplication openURL:.
+// Handing an unresolved Teak universal link to UIApplication openURL: does NOT
+// re-enter the app — iOS routes the https URL to Safari. Shown here as the
+// counter-example to the helper below.
 - (IBAction)deferredDeepLinkOpenURL {
   NSURL* url = [NSURL URLWithString:@"https://teak-dev2.freechips.link/1bafc0c4f1"];
-  NSLog(@"C-744 path A openURL: %@", url);
+  NSLog(@"deferredDeepLinkOpenURL: %@", url);
   [[UIApplication sharedApplication] openURL:url options:@{} completionHandler:^(BOOL success) {
-    NSLog(@"C-744 path A openURL completion success=%d", success);
+    NSLog(@"deferredDeepLinkOpenURL completion success=%d", success);
   }];
 }
 
-// C-744 test path B: the public helper. This is what a customer calls when an
-// attribution SDK (e.g. Singular) hands them an unresolved universal link after
-// launch — no bridging-header forward decl, no NSUserActivity knowledge needed.
+// When an attribution SDK (e.g. Singular) hands you an unresolved Teak universal
+// link after launch, processDeferredDeepLink: resolves it as if the user had
+// tapped it: it starts a session, claims any reward, and routes the deep link.
 - (IBAction)deferredDeepLinkProcessHelper {
   NSURL* url = [NSURL URLWithString:@"https://teak-dev2.freechips.link/1bafc0c4f1"];
-  NSLog(@"C-744 path B processDeferredDeepLink: %@", url);
+  NSLog(@"deferredDeepLinkProcessHelper: %@", url);
   [[Teak sharedInstance] processDeferredDeepLink:url];
 }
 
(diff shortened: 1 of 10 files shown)


Sample/Sample/ViewController.m
```diff
@@ -99,6 +99,26 @@ - (void)paymentQueue:(SKPaymentQueue*)queue updatedTransactions:(NSArray<SKPayme
   }
 }
 
+// Singular hands an unresolved Teak universal link to the app post-launch and we
+// need it to start a session, claim the reward, and route the deep link. C-744
+// test path A: hand the URL to UIApplication openURL:.
+- (IBAction)deferredDeepLinkOpenURL {
+  NSURL* url = [NSURL URLWithString:@"https://teak-dev2.freechips.link/1bafc0c4f1"];
+  NSLog(@"C-744 path A openURL: %@", url);
```
AlexSc: Make comments/log messages/etc timeless

Teak/Teak.m
```diff
@@ -1193,4 +1193,22 @@ - (BOOL)application:(UIApplication*)application continueUserActivity:(NSUserActi
   return YES;
 }
 
+- (void)processDeferredDeepLink:(NSURL*)url {
```
AlexSc, line 1199: Instead of implementing the helper in terms of "call this baroque internal lifecycle thing", put the URL resolution in the helper and have the lifecycle method use the same method

AlexSc (changes requested):

Teak/Teak.m
TeakBuilds: [sdks-teak-ios-worker-c-744] Done in 20b5040. processDeferredDeepLink: now builds the launch data straight from the URL via fromUniversalLink: — no more manufactured NSUserActivity / nil-application round-trip. Extracted a private handleLaunchData: sink that both the helper and continueUserActivity: funnel through; continueUserActivity: keeps calling fromUserActivity: so the browsing-web and live-activity guards stay intact. Zero behavior change on the production path (fromUserActivity: already delegates browsing-web links to fromUniversalLink:). fastlane 129/0.

Sample/Sample/ViewController.m
TeakBuilds: [sdks-teak-ios-worker-c-744] Done in 487730c. Swept the whole Sample deferred-deep-link diff, not just this line: comments and NSLog strings now describe the behavior (openURL routes to Safari and does not re-enter the app; processDeferredDeepLink: resolves the link as if tapped) instead of referencing the ticket or the path A/B investigation framing.

AlexSc (approved):

PR #151, opened 2026-06-29
Pin NSDateFormatter locale to en_US_POSIX (C-302)

TeakBuilds:
https://linear.app/teakio/issue/C-302/purchase-time-is-being-localized-and-it-should-not-be

NSDateFormatter defaults to [NSLocale currentLocale], so Arabic-locale devices produced Eastern-Arabic numerals in date strings sent to the server → ArgumentError: mon out of range.

Fix: add `[formatter setLocale:[NSLocale localeWithLocaleIdentifier:@"en_US_POSIX"]]` to each formatter that sends timestamps to an external system. Also caught the same bug in TeakRaven.m (Sentry timestamps) during a sibling audit — same one-line fix. TeakLog.m was already clean.

New `SKPaymentObserverTests.m` has two regression tests verifying both formatter configurations produce ASCII ISO 8601 output.

[sdks-teak-ios-worker-c-302]


AlexSc (approved):
Followup issue to centralize a singleton correctly configured NSDateFormatter so we just keep this sorted

PR #155, opened 2026-06-30
Retry requests when the OS closes a socket without reopening it

### LINKED ISSUE #85: Retry requests that fail because the OS closed or did not reopen the socket
This can happen when an app comes back from background

TeakBuilds:
Closes #85
Linear: https://linear.app/teakio/issue/C-386

`TeakRequest`'s `response:payload:withError:` discarded the NSError entirely (`TeakUnused(error)`) — a transport-level failure (no HTTP response) just fired the callback with an empty payload, no retry. The known cause is the OS closing a pooled connection's socket while backgrounded and failing to reopen it, surfacing `NSPOSIXErrorDomain`/`ECONNABORTED` (53) at the top level of the task error. `TeakLog.m` already retries this exact failure on its log-sending path; this ports the same fix onto the main request path (purchases, rewards, identify, `/me/events`, etc).

- `+[TeakRequest isRetryableSocketError:]` — pure predicate, checks top-level domain/code and `NSUnderlyingErrorKey` for a nested POSIX/53.
- Scoped to 53 only (not `NSURLErrorNetworkConnectionLost`/-1005): 53 fires before the request body is sent (stale connection abort), so retry can't double-submit; -1005 is a mid-load severance that may have already reached the server.
- One-shot retry via the existing `[self send]` path, 1.5s delay (matches TeakLog's), guarded by a new `retriedAfterSocketError` flag.
- New logs close the observability gap the swallowed error left: `request.retry.socket_error` (retry fires) / `request.retry.socket_error.exhausted` (gives up — only on the true give-up path, not when a server-configured 5xx retry picks it up instead).

## Test plan
- [x] `bundle exec fastlane test` — 146/146 green
- [x] New `TeakRequestSocketErrorTests.m` exercises the predicate directly (nil, top-level match/mismatch, cross-domain same-code boundary, -1005 boundary, nested match/mismatch)

### DIFF
--- a/Automated/AutomatedTests/TeakRequestSocketErrorTests.m
+++ b/Automated/AutomatedTests/TeakRequestSocketErrorTests.m
@@ -1,0 +1,77 @@
+#import <XCTest/XCTest.h>
+
+#import "TeakRequest.h"
+#import <Teak/Teak.h>
+
+#include <sys/errno.h>
+
+@import OCHamcrest;
+@import OCMockito;
+
+// Re-expose the internal predicate so tests can exercise it without driving
+// a full NSURLSession round-trip. See CLAUDE.md "Header imports in tests".
+@interface TeakRequest (SocketErrorTests)
++ (BOOL)isRetryableSocketError:(NSError* _Nullable)error;
+@end
+
+@interface TeakRequestSocketErrorTests : XCTestCase
+@end
+
+@implementation TeakRequestSocketErrorTests
+
+#pragma mark - nil
+
+- (void)testNilErrorIsNotRetryable {
+  XCTAssertFalse([TeakRequest isRetryableSocketError:nil]);
+}
+
+#pragma mark - The verified production shape: top-level NSPOSIXErrorDomain/ECONNABORTED
+
+// This is the exact error TeakLog's analogous retry was built from
+// production telemetry to handle, and matches Apple's own documented
+// "NSURLSessionTask receives error 53 when app is relaunched from
+// background" failure mode.
+- (void)testTopLevelPOSIXConnectionAbortedIsRetryable {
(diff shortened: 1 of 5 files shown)


AlexSc (changes requested):
Instead of a bool "did I already retry" let's make it a count of retries now. We can still abort additional retries after one shot, but let's decouple call site handling from our business logic on when to stop retrying now.

Automated/AutomatedTests/TeakRequestSocketErrorTests.m
```diff
@@ -0,0 +1,116 @@
...
+
+- (void)testDoesNotRetryASocketErrorAtTheLimit {
+  NSError* error = [NSError errorWithDomain:NSPOSIXErrorDomain code:ECONNABORTED userInfo:nil];
+  XCTAssertFalse([TeakRequest shouldRetrySocketError:error retryCount:1]);
+}
+
+- (void)testDoesNotRetryASocketErrorPastTheLimit {
+  NSError* error = [NSError errorWithDomain:NSPOSIXErrorDomain code:ECONNABORTED userInfo:nil];
+  XCTAssertFalse([TeakRequest shouldRetrySocketError:error retryCount:2]);
+}
```
AlexSc, line 106: Rework these to use the constant

AlexSc (approved):
Test nits must be cleaned before merge, but no need to rereview. Fix and merge on CI pass

Automated/AutomatedTests/TeakRequestSocketErrorTests.m
TeakBuilds: [sdks-teak-ios-worker-c-386] Fixed in 7301c54 — `TeakRequestMaxSocketRetries` now exposed via `TeakRequest+Internal.h`, and these assert `TeakRequestMaxSocketRetries - 1` (retries) / `TeakRequestMaxSocketRetries` and `+ 1` (stop) instead of hardcoded counts.

PR #163, opened 2026-07-06
Serialize TeakUserProfile attribute-dict access on the operationQueue (C-955)

TeakBuilds:
Linear: https://linear.app/teakio/issue/C-955

`TeakUserProfile setAttribute:forKey:inDictionary:` read the attribute dict (nil-guard) and stamped `firstSetTime` on the caller's default-qos queue while the write ran on the serial `io.teak.sdk.operationQueue`. Concurrent read/write of the same `NSMutableDictionary` — the serial writer rehashing/freeing an evicted value while a caller-queue reader retains it — is the use-after-free behind Crash Family 2 (~2.4K prod crashes across Quick Hit / Jackpot Party / 88 Fortunes / Hot Shot).

**Fix:** move the nil-guard, `firstSetTime` stamp, equality test, and write all inside the existing `dispatch_async([Teak operationQueue], …)` block, finishing what 47b97dd started. Once the guard-read and the write both run on the serial queue they're atomic w.r.t. each other — no reader observes the dict mid-mutation. `firstSetTime` stays right after the nil-guard and before the equality check, so `ms_since_first_event` semantics are unchanged.

**Test:** `TeakUserProfileAttributeQueueTests` follows the `TeakSessionDeferredCallbackTests` precedent — a deterministic invariant proof, not a flaky TSan repro. Blocks the serial queue behind a semaphore, calls the real setters, and asserts `firstSetTime` is still nil (deferred behind the gate). Verified it fails on both string + numeric paths when the fix is reverted.


AlexSc (approved):
Lead, I also want a followup issue (we can try it before TSan) to have a *fable* model agent review the codebase for potential race conditions like this

PR #166, opened 2026-07-06
Add ThreadSanitizer race guard and race-testing methodology doc

TeakBuilds:
Stands up ThreadSanitizer as a race-detection guard for the SDK — run per-commit for visibility and wired as a release gate on tag — plus a methodology doc for writing race tests. **No production fixes.**

Linear: https://linear.app/teakio/issue/C-961

## What's here

- `Automated/AutomatedTests/RaceTripwireTests.m` — a race-detection regression guard driving a real Teak type under ThreadSanitizer.
- `Automated/RACE_TESTING.md` — how to test a concurrency bug in this SDK and prove the test is trustworthy (linked from CLAUDE.md's Adding Tests section).
- A `test_race` fastlane lane that runs the guard under TSan: per-commit in `un-tagged-build` for visibility, and as a release gate in `tagged-build` (a red run blocks the deploy).
- `Automated/tsan-suppressions.txt` — vendored-code-only suppressions (empty for now), deliberately outside `Teak/`.

## The guard

`testUserProfileAttributeDictIsSerializedUnderConcurrency` drives the real attribute setter from two threads under TSan. The serialization fix (hopping every access onto the serial operationQueue) keeps it green; reverting the fix makes TSan report `race on NSMutableDictionary` in `-[__NSDictionaryM setObject:forKeyedSubscript:]`. It exercises production code and fails when the fix is reverted — a real regression guard, verified both ways.

The fast `test` lane skips this class (it carries signal only under a sanitizer); `test_race` carries the TSan run separately, so no double-run.

## Methodology doc

`Automated/RACE_TESTING.md` captures what standing this up taught us. Classify the shared state first (immortal / freeable-pointer / mutable-container), because that decides which detector fires: ThreadSanitizer for both-sides-instrumented races, a dynamic CoreFoundation over-release crash repro for the nonatomic-strong UAF that TSan is **blind** to (the setter's release lives in uninstrumented `objc_storeStrong`, so TSan never pairs it with the reader's load), a static assertion for immortal slots. It includes a buildable crash-repro skeleton, the "revert check" that makes any race test trustworthy, why `atomic` closes the load→retain gap, and the standing workflow — each fix ships its own red→green test in its fix PR, so this branch stays green.

## Why no standing reds

An earlier revision carried two red guards for still-open defects (a nonatomic-strong session-property UAF and a shared raven report dict). Per review, standing reds don't belong on this branch — each fix ships its own red→green test in its fix PR — so those were stripped and their tests move to their fix PRs. `test_race` is now all-green.

## Wiring notes

- **Per-commit visibility + tag-build release gate**: `test_race` runs per-commit in `un-tagged-build` (non-gating there) and again in `tagged-build`, where `deploy_versioned` and `hold` require it, so a red guard blocks the release deploy chain. This supersedes the original nightly cron.
- **halt_on_error=0**: the lane sets TSan `halt_on_error=0` explicitly so it reports every race without aborting on the first. `store_test_results`/`store_artifacts` run `when: always`, so results upload on a red too.
- **Suppressions**: vendored-only and outside `Teak/`, so a suppression can never be mistaken for a blessed race in first-party code.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,6 +22,35 @@ jobs:
           command: bundle exec fastlane test
       - store_test_results:
           path: test_output/automated
+  test_race:
+    parameters:
+      xcode-version:
+        type: string
+        default: "14.3.1"
+    macos:
+      xcode: << parameters.xcode-version >>
+    resource_class: m4pro.medium
+    shell: /bin/bash --login -eo pipefail
+    steps:
+      - ios_setup
+      - run:
+          name: Run Race Tripwires (ThreadSanitizer)
+          command: |
+            mkdir -p test_output/race
+            # The lane runs every tripwire and exits nonzero if any is red; swallow that here so the
+            # results are always stored, then re-surface it as a job failure in a later step.
+            bundle exec fastlane test_race || touch test_output/race/.failed
+      - store_test_results:
+          path: test_output/race
+      - store_artifacts:
+          path: test_output/race
+      - run:
+          name: Surface race tripwire result
+          command: |
+            if [ -f test_output/race/.failed ]; then
+              echo "Race tripwires reported failures — a guarded defect is still open (see stored results)."
+              exit 1
+            fi
   build:
     parameters:
       xcode-version:
@@ -176,6 +205,19 @@ workflows:
           context:
             - iOS-Build
       - build:
+          context:
+            - iOS-Build
+
+  nightly-race:
+    triggers:
+      - schedule:
+          cron: "0 0 * * 1-5"
+          filters:
+            branches:
+              only:
+                - 4.3-stable
+    jobs:
+      - test_race:
           context:
             - iOS-Build
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,24 +33,17 @@ jobs:
     shell: /bin/bash --login -eo pipefail
     steps:
       - ios_setup
+      # A red run means a guarded defect is still open: the lane exits nonzero and fails the job.
+      # store_* run `when: always` so the results and artifacts still upload on that red.
       - run:
           name: Run Race Tripwires (ThreadSanitizer)
-          command: |
-            mkdir -p test_output/race
-            # The lane runs every tripwire and exits nonzero if any is red; swallow that here so the
-            # results are always stored, then re-surface it as a job failure in a later step.
-            bundle exec fastlane test_race || touch test_output/race/.failed
+          command: bundle exec fastlane test_race
       - store_test_results:
           path: test_output/race
+          when: always
       - store_artifacts:
           path: test_output/race
-      - run:
-          name: Surface race tripwire result
-          command: |
-            if [ -f test_output/race/.failed ]; then
-              echo "Race tripwires reported failures — a guarded defect is still open (see stored results)."
-              exit 1
-            fi
+          when: always
   build:
     parameters:
       xcode-version:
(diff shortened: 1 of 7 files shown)


AlexSc (changes requested):
I'd like to see these run in CI before I review. Push a commit to this branch that runs the new tests on every commit. Once that's done I'll review, and I may make the call to run tsan on every commit and eat the reds until the existing issues are resolved (update them to high/urgent priority)

AlexSc (changes requested):
Drop references to issues or PR description in the code. Codebase must be standalone.

Cut tests that don't assert on production behavior. You can leave commentary indicating what we've confirmed is testable under TSan.

The goal here is to actually improve our testing posture, not just performative work.

.circleci/config.yml
```diff
@@ -22,6 +22,35 @@ jobs:
...
+      xcode: << parameters.xcode-version >>
+    resource_class: m4pro.medium
+    shell: /bin/bash --login -eo pipefail
+    steps:
+      - ios_setup
+      - run:
+          name: Run Race Tripwires (ThreadSanitizer)
+          command: |
+            mkdir -p test_output/race
+            # The lane runs every tripwire and exits nonzero if any is red; swallow that here so the
```
AlexSc: Why are we swallowing and resurfacing the failure?

.circleci/config.yml
```diff
@@ -187,6 +216,16 @@ workflows:
           filters:
             tags:
               ignore: /.*/
+      # Runs the race tripwires under TSan on every commit. Expected red while the guarded defects
+      # are open — two tripwires assert an as-yet-unfixed invariant and fail by design, and that red
+      # is the tracking signal, not a regression. Intentionally not a dependency of tag_build, so
+      # the known reds don't freeze the release/tag-promote flow.
```
AlexSc: Run on tag. We are going to fix any red before cutting 4.3.14, guaranteed. Don't leave footguns laying around.

Automated/AutomatedTests/RaceTripwireTests.m
```diff
@@ -0,0 +1,157 @@
...
+
+  // Drain the queued setter blocks so none outlive the test.
+  dispatch_sync([Teak operationQueue], ^{
+  });
+}
+
+// Session strong-property atomicity guard (property introspection). The session reply block
+// reassigns these strong properties on one queue while the duration/report path reads and uses them
+// on another; nonatomic strong reassignment releases the prior object mid-read (use-after-free).
+// Making them atomic hands the reader a retained snapshot. Red while they remain nonatomic.
```
AlexSc: ... making them atomic does what now? I was under the impression that the atomicity was mostly a red herring under arm64 and the real fix is capturing to local strong references instead of pulling from the shared state object

AlexSc (changes requested):

.circleci/config.yml
TeakBuilds: [sdks-worker-c-961] Fixed in bf3c810 — the swallow/resurface only existed to keep store_test_results running after a failed step. Dropped it: `fastlane test_race` fails the job directly, and store_test_results + store_artifacts are now `when: always`, so results/artifacts still upload on a red.

.circleci/config.yml
TeakBuilds: [sdks-worker-c-961] Done in aeb81fb — test_race now gates the release: added it to the tagged-build workflow with deploy_versioned + hold requiring it, so a red tripwire blocks the version deploy (and the whole deploy chain). Kept the per-commit run in un-tagged-build for visibility (non-gating there, so a red does not block promote-tag creation). Put the gate on both deploy_versioned and hold, since deploy_latest chains off hold — otherwise `latest` could slip past a red.

AlexSc (approved):

PR #168, opened 2026-07-07
Make TeakSession serverSessionId/countryCode/userProfile atomic (UAF fix)

TeakBuilds:
Fixes a cross-thread use-after-free: `TeakSession.serverSessionId`, `countryCode`, and `userProfile` were `nonatomic` strong properties reassigned in the identify-reply request callback while read concurrently by the heartbeat queue and the duration-report background block. A nonatomic setter compiles to a bare `objc_storeStrong` with no synchronization, so a concurrent reader can retain a pointer mid-free.

Linear: https://linear.app/teakio/issue/C-965

## Changes

- `TeakSession.h`/`.m`: all three properties → `atomic` (matches the existing `currentState`/`reportDurationBlock` pattern).
- Capture-to-local at every multi-read site, since atomicity alone doesn't protect a nil-check-then-use pair from seeing the value change in between:
  - `sendHeartbeat` — single-read `countryCode` before the URL-building ternary.
  - The `Session_resume` payload builder (on the `Expiring` KVO transition) — single-read `serverSessionId`.
  - The `userProfile` send-dispatch site — replaced a `weakSelf`/`blockSelf` dance with a direct local capture; `TeakUserProfile` holds a *strong* `session` back-reference (`TeakRequest+Internal.h`), so `TeakSession` can't dealloc while it holds a live `userProfile` regardless of the weak-self dance — removing it is behavior-preserving, not just a simplification.
  - `reportDurationBlock`'s body — added a `serverSessionId` local capture alongside the existing cancellation-check capture.
- `TeakSessionLockingTests.m`: static atomic-declaration tests (`isNonatomicProperty:` introspection) for all three properties, as a cheap permanent guard on top of the dynamic repro below.
- `RaceTripwireTests.m`: three independent dynamic crash-repro tests (CF over-release trap, per `Automated/RACE_TESTING.md` §3) — one per property, not combined, so a revert of any single property's fix is independently attributable to a red result.

## Revert checks (observed, not asserted)

Each of the three dynamic tests was reverted and restored independently and watched crash/pass:

- `testServerSessionIdIsAtomicUnderConcurrency` — crashes reliably reverted to `nonatomic` (200k iterations, fresh >15-char heap `NSString`s, `.length` dereference on the reader); clean restored to `atomic`. Other two tests stayed green throughout.
- `testCountryCodeIsAtomicUnderConcurrency` — same pattern and result, independently.
- `testUserProfileIsAtomicUnderConcurrency` — needed more than the other two's baseline to reproduce reliably: `userProfile`'s pointee is a plain object, not a `CFString`, so touching only `.stringAttributes.count` was a false green (freed memory was often reused by a same-shaped allocation before the read landed). Bumped to 2,000,000 iterations and added an isa-touching dereference (`NSStringFromClass([p class]).length`) — confirmed crashing reliably reverted (3 runs) and clean restored (2 runs, one under `MallocScribble`/`MallocPreScribble`).

Full test suite (`bundle exec fastlane test`) and the race lane (`bundle exec fastlane test_race`) both green with the fix in place.

### DIFF
--- a/Automated/AutomatedTests/RaceTripwireTests.m
+++ b/Automated/AutomatedTests/RaceTripwireTests.m
@@ -9,10 +9,11 @@
 // detector can observe, so a red run means a real regression — the fix it guards was reverted. Two
 // kinds of guard live here: deterministic CF-over-release crash repros (serverSessionId/countryCode/
 // userProfile) that carry signal on their own, and a ThreadSanitizer-only serialization guard (the
-// attribute-dict test) that needs the sanitizer to see the race at all. The fast per-commit `test`
-// lane skips the whole class — the crash repros are too slow (up to 2M iterations) for a quick
-// per-commit signal, and the serialization test needs TSan anyway; the dedicated `test_race` lane
-// runs all of them, with ThreadSanitizer attached for the one test that needs it.
+// attribute-dict test) that needs the sanitizer to see the race at all. Both run every commit, just
+// on different lanes: the `test` lane skips this whole class to keep the fast per-commit signal
+// quick — the crash repros are too slow (up to 2M iterations) for it, and the serialization test
+// needs TSan anyway. The `test_race` lane runs all of them, every commit, with ThreadSanitizer
+// attached for the one test that needs it, and additionally gates tagged-build releases.
 //
 // See Automated/RACE_TESTING.md for the race-testing methodology — which detector catches which
 // race class, and why some classes (e.g. lock-inversion deadlocks) get no in-process guard here.
--- a/Automated/AutomatedTests/RaceTripwireTests.m
+++ b/Automated/AutomatedTests/RaceTripwireTests.m
@@ -6,10 +6,13 @@
 #import <Teak/Teak.h>
 
 // Race-detection regression guard. Drives a real Teak type under concurrency and pins an invariant a
-// detector can observe, so a red run means a real regression — the fix it guards was reverted. The
-// fast per-commit `test` lane skips this class (it carries signal only under a sanitizer); the
-// `test_race` lane runs it under ThreadSanitizer, where the race resurfaces if the serialization fix
-// is removed.
+// detector can observe, so a red run means a real regression — the fix it guards was reverted. Two
+// kinds of guard live here: deterministic CF-over-release crash repros (serverSessionId/countryCode/
+// userProfile) that carry signal on their own, and a ThreadSanitizer-only serialization guard (the
+// attribute-dict test) that needs the sanitizer to see the race at all. The fast per-commit `test`
+// lane skips the whole class — the crash repros are too slow (up to 2M iterations) for a quick
+// per-commit signal, and the serialization test needs TSan anyway; the dedicated `test_race` lane
+// runs all of them, with ThreadSanitizer attached for the one test that needs it.
 //
 // See Automated/RACE_TESTING.md for the race-testing methodology — which detector catches which
 // race class, and why some classes (e.g. lock-inversion deadlocks) get no in-process guard here.
--- a/Automated/AutomatedTests/RaceTripwireTests.m
+++ b/Automated/AutomatedTests/RaceTripwireTests.m
@@ -1,6 +1,7 @@
 #import <XCTest/XCTest.h>
 #import <stdatomic.h>
 
+#import "TeakSession.h"
 #import "TeakUserProfile.h"
 #import <Teak/Teak.h>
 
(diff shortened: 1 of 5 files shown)


Automated/AutomatedTests/RaceTripwireTests.m
```diff
@@ -21,6 +25,16 @@ @interface TeakUserProfile (RaceTripwire)
 @property (strong, nonatomic) NSMutableDictionary* stringAttributes;
 @end
 
+// countryCode/serverSessionId are private (declared only in TeakSession.m's class extension);
+// userProfile is public but readonly there. Re-declared here for full read-write access, per the
+// project convention of redeclaring internal properties in the test file rather than importing
+// Teak+Internal.h.
```
AlexSc, line 32: Let's file a follow up issue, next patch, to switch to using private includes instead of redecl's in test

Automated/AutomatedTests/RaceTripwireTests.m
```diff
@@ -1,14 +1,18 @@
...
 // Race-detection regression guard. Drives a real Teak type under concurrency and pins an invariant a
-// detector can observe, so a red run means a real regression — the fix it guards was reverted. The
-// fast per-commit `test` lane skips this class (it carries signal only under a sanitizer); the
-// `test_race` lane runs it under ThreadSanitizer, where the race resurfaces if the serialization fix
-// is removed.
+// detector can observe, so a red run means a real regression — the fix it guards was reverted. Two
+// kinds of guard live here: deterministic CF-over-release crash repros (serverSessionId/countryCode/
+// userProfile) that carry signal on their own, and a ThreadSanitizer-only serialization guard (the
+// attribute-dict test) that needs the sanitizer to see the race at all. The fast per-commit `test`
+// lane skips the whole class — the crash repros are too slow (up to 2M iterations) for a quick
```
AlexSc: We are running per-commit, so this comment is inaccurate.

Teak/TeakSession.m
```diff
@@ -846,12 +857,14 @@ + (void)checkLaunchDataForDeepLinkAndDispatchEvents:(nonnull TeakLaunchData*)lau
...
 
-      // Send user profile out now
-      if (self.userProfile != nil) {
-        __weak typeof(self) weakSelf = self;
+      // Send user profile out now. Single read: capture the profile that's live now rather than
+      // re-reading self.userProfile when the async block runs, which could be a different (or nil)
+      // value by then. No weakSelf/blockSelf needed — userProfile.session strongly retains this
+      // session (TeakRequest+Internal), so self can't be deallocated while userProfile is live.
+      TeakUserProfile* userProfile = self.userProfile;
+      if (userProfile != nil) {
```
AlexSc, line 865: Could we end up in a situation where a previous user profile is still processing its send when a newer one also enters this path?

Are there options to serialize sends, or preempt sends?

I suspect the answer is "Yes we could end up in that situation, and so can the new code", if that's true I'd say follow up next patch, if the new code introduces a potential race there then I'd like to resolve it this patch.

AlexSc (changes requested):

Automated/AutomatedTests/RaceTripwireTests.m
TeakBuilds: [sdks-apm] filed as C-978 (SDK next patch): https://linear.app/teakio/issue/C-978/use-private-test-includes-instead-of-property-redeclarations-in-race

Automated/AutomatedTests/RaceTripwireTests.m
TeakBuilds: [sdks-worker-c-965] Good catch — fixed. `test_race` runs every commit too (non-gating in `un-tagged-build`, alongside `test`; it additionally gates deploys in `tagged-build`). Reworded so the comment doesn't imply otherwise.

Teak/TeakSession.m
TeakBuilds: [sdks-worker-c-965] Traced it: yes, this is possible. Two lines above this hunk, the pre-existing comment "Stop heartbeat, Expiring->Expiring is possible, so no invalid data here" confirms this whole block can re-run before an earlier send finishes — `TeakRequest.send` (TeakRequest.m:312) builds the request, calls `[dataTask resume]`, and returns immediately; it doesn't wait for the response. So two `userProfile send` calls from consecutive Expiring entries can have their network requests genuinely overlapping in flight.

This is pre-existing, not something this patch changes. The old weakSelf/blockSelf code dispatched the exact same one-async-block-per-Expiring-entry structure with no serialization or preemption between them; my change only affects *which* profile reference gets captured (dispatch-time snapshot vs. re-read at execution time), not the dispatch cardinality or TeakRequest's fire-and-forget send. Also worth noting: TeakRequest already tracks in-flight requests by task identifier under a mutex (`requestsInFlight`, TeakRequest.m:322-324), so overlapping sends aren't a memory-safety race — at most a duplicate/out-of-order network call.

Agreed with your suspicion — I'd defer serialize-or-preempt to a follow-up rather than fold it into this atomicity-scoped patch.

TeakBuilds: [sdks-apm] pre-existing + not memory-safety → deferring to C-979 next patch, this PR only changes which snapshot is captured, not dispatch cardinality: https://linear.app/teakio/issue/C-979/serialize-or-preempt-overlapping-userprofile-sends-on-session-expiry

AlexSc (approved):

PR #169, opened 2026-07-07
Make TeakPushState.stateChain atomic to fix a cross-thread use-after-free

TeakBuilds:
https://linear.app/teakio/issue/C-972/teakpushstatestatechain-cow-swap-read-lock-free-u6

## Summary
- `stateChain` is written under `@synchronized(self)` in `updateCurrentState:` (COW discipline already in place — the writer swaps in a whole new array) but read lock-free by `cachedPushState` and `serializedStateChain`. That's the same nonatomic-strong freeable-pointee UAF class already fixed for `TeakSession`'s `countryCode`/`userProfile`/`serverSessionId` (#166) and Raven's user dict (#167). Flipped `stateChain` to `atomic`.
- Both lock-free read sites dereference the property exactly once, so no capture-to-local was needed beyond the atomic accessor.
- Added a static atomic-declaration test (`TeakPushStateLockingTests.m`) and a dynamic crash-repro test (`RaceTripwireTests.m`), per `Automated/RACE_TESTING.md` §3. The array's pointee is a plain object (not a CFString), so the repro uses the isa-touch technique + a higher iteration count (2M), same as `userProfile`'s.

## Revert-check
Reverted `stateChain` to `nonatomic` and confirmed both new tests go red (the dynamic repro crashes via the CF over-release trap; the static test fails the runtime introspection assertion). Restored `atomic` and confirmed both go green.

## Test plan
- [x] `bundle exec fastlane test` — 163 tests, 0 failures
- [x] Targeted run of `RaceTripwireTests` (7 tests) — 0 failures
- [x] Revert-check: red on nonatomic, green on atomic, for both new tests

### DIFF
--- a/Automated/AutomatedTests/TeakPushStateLockingTests.m
+++ b/Automated/AutomatedTests/TeakPushStateLockingTests.m
@@ -1,0 +1,34 @@
+#import <XCTest/XCTest.h>
+
+#import "TeakPushState.h"
+#import <objc/runtime.h>
+
+// TeakPushState.stateChain is written under @synchronized(self) (updateCurrentState:) but read
+// lock-free by cachedPushState and serializedStateChain. COW discipline means the writer swaps in
+// a brand-new array rather than mutating in place, so the old array is freed on swap — a
+// nonatomic-strong reassignment on one thread races a read+retain on another, retaining a pointer
+// that's being freed underneath it. See RaceTripwireTests.m for the dynamic use-after-free repro
+// backing this. This test is the same cheap permanent static pin used for TeakSession's
+// atomic properties in TeakSessionLockingTests.m — it isn't the guard that matters (the dynamic
+// repro is), but it catches an accidental revert before the slower dynamic lane even runs.
+@interface TeakPushStateLockingTests : XCTestCase
+@end
+
+@implementation TeakPushStateLockingTests
+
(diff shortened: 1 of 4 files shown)


Automated/AutomatedTests/TeakPushStateLockingTests.m
```diff
@@ -0,0 +1,34 @@
...
+// atomic properties in TeakSessionLockingTests.m — it isn't the guard that matters (the dynamic
+// repro is), but it catches an accidental revert before the slower dynamic lane even runs.
+@interface TeakPushStateLockingTests : XCTestCase
+@end
+
+@implementation TeakPushStateLockingTests
+
+// YES if the named property carries the nonatomic flag (the "N" token in its runtime
+// attribute string), which would reintroduce the torn-read race.
+- (BOOL)isNonatomicProperty:(const char*)name {
```
AlexSc, line 21: Follow-up refactoring, extract this to a shared helper, change to isAtomicProperty, and assert true.

AlexSc (approved):

Automated/AutomatedTests/TeakPushStateLockingTests.m
TeakBuilds: [sdks-worker-c-972] Agreed, filing as a follow-up rather than blocking this PR — flagged in the issue channel for lead-pm to route.

PR #172, opened 2026-07-07
Synchronize TeakLink's route registry against concurrent mutation and enumeration

TeakBuilds:
Fixes a mutate-while-enumerate race in TeakLink's static route registry (C-969, from the C-962 audit, U3).

https://linear.app/teakio/issue/C-969/teaklink-static-route-registry-mutate-while-enumerate-u3

## Problem
`registerRoute` writes the static route-registry `NSMutableDictionary` from any host thread with no threading contract, while `handleDeepLink` and `routeNamesAndDescriptions` enumerate it on the op queue with no synchronization. A host registering routes lazily post-launch while the launch deep link resolves concurrently races a mutation against an enumeration — `NSGenericException` or a rehash-UAF.

## Fix
`deepLinkRegistration`'s existing `dispatch_once` already makes it a stable, create-once, never-reassigned `@synchronized` token. Wrap the write in `registerRoute` and take a copy-then-enumerate snapshot in `handleDeepLink`/`routeNamesAndDescriptions` (factored into a shared `+deepLinkRegistrationSnapshot` helper), all under that lock — matches the `TeakIntegrationChecker.m` precedent of synchronizing directly on the mutable container instance.

## Test
Mutable-container race per `Automated/RACE_TESTING.md` → ThreadSanitizer is the detector. New `testTeakLinkRouteRegistryIsSerializedUnderConcurrency` drives the real `registerRoute`/`handleDeepLink`/`routeNamesAndDescriptions` concurrently. Revert check (isolated runs, off the shared CI-adjacent simulator lane): **red** with the `@synchronized` removed — TSan reports `race on NSMutableDictionary` between `registerRoute`'s `setObject:forKey:` and `handleDeepLink`'s `countByEnumeratingWithState:` — **green** with the fix restored. Full non-TSan `test` lane also green (161/161).

## Changelog (proposed, for the consolidated 4.3.14 entry)
Fixed a rare crash/data race when registering a TeakLink deep link route concurrently with deep link resolution at launch.

### DIFF
--- a/Automated/AutomatedTests/RaceTripwireTests.m
+++ b/Automated/AutomatedTests/RaceTripwireTests.m
@@ -4,6 +4,7 @@
 #import "TeakAppConfiguration.h"
 #import "TeakConfiguration.h"
 #import "TeakDeviceConfiguration.h"
+#import "TeakLink.h"
 #import "TeakLog.h"
 #import "TeakRaven.h"
 #import "TeakSession.h"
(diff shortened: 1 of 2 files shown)


AlexSc (approved):
I was going to ask why the payment queue didn't synchronize on its array object but figured there was something up.

However this is the same basic pattern and it does synchronize on the underlying container. I prefer this. follow-up to go back to the payment observer and synchronize on the container instead of a string "mutex"

PR #173, opened 2026-07-07
Serialize TeakSession deviceConfiguration KVO observers on a dedicated lock, detach at replacement

TeakBuilds:
Corrects the observer lifecycle for the process-shared `deviceConfiguration` KVO, and makes the add/remove serialization verifiable from a single lock.

## Background

Every `TeakSession` observes `advertisingIdentifier`/`pushToken`/`liveActivityPushToStartToken` on the singleton `deviceConfiguration`, adding those observers in `-init`. Two problems:

1. **Observers are never removed.** They were only unregistered in `-dealloc` — but a session retains itself through the shared event-handler registry (`TeakEventHandlerSet`, removed only in `-dealloc`), a retain cycle, so `-dealloc` never runs. The observers accumulate on the shared object for the life of the process.
2. **Serialization rested on an unstated call-site invariant.** KVO add/remove is not thread-safe (Swift **SR-6795** is exactly this observation-info corruption). Every construct and teardown happens under `currentSessionMutex` today, so add/remove *is* serialized — but only because every call site happens to hold that lock, which a reader can't verify locally.

## Fix

- **One dedicated leaf lock** (`deviceConfigurationObserverMutex`) is held directly across `-init`'s three `addObserver` calls **and** `-detachDeviceConfigurationObservers`' three `removeObserver` calls. Serialization is now verifiable by grepping that one lock's use sites, independent of the call-site invariant — the readability the review asked for. It's belt-and-suspenders for today's live paths (all under `currentSessionMutex`) and becomes the sole guard once `-dealloc` runs (below).
- **Detach at replacement bounds the growth.** The three replacement sites (`logoutReusingCurrentSession:`, `didLaunchWithData:`, `+currentSession`) call `-detachDeviceConfigurationObservers`, so the outgoing session's observers come off the shared object deterministically instead of accumulating. A flag makes removal idempotent.
- **Dedicated token, not `@synchronized(self)`**: `self` is entangled with `currentSessionMutex` (the `currentState` handler runs under `@synchronized(self)`), so routing KVO through `self` would couple it to that ordering. The token is a true leaf — `addObserver`/`removeObserver` use options `New|Old` (never `Initial`), fire no synchronous `observeValueForKeyPath:`, so nothing that takes `self`/`currentSessionMutex` runs while it's held.
- **Kept-but-inert `-dealloc` fallback.** The `-dealloc` detach can't run today (immortal sessions), but it's retained: the moment the self-retention leak is fixed and sessions become mortal, `-dealloc` runs on a background queue *off* `currentSessionMutex`, and its `removeObserver` would race a new session's `-init` `addObserver` — the leaf lock is what serializes that. The "dealloc races init" race is real but **dormant**; the fallback + leaf lock are its forward-defense. (Leak tracked separately as C-981.)
- **Keeps** logout's inner `@synchronized(currentSession)` (a prior revision had removed it): it holds the outgoing session's monitor across the `Expiring`→`Expired` pair, blocking an in-flight `hostname` KVO callback (`-setState:Configured`) from interleaving between the two `-setState:` calls. `Expiring`→`Configured` is legal; the following `Expired` would then be rejected (`Configured` has no `Expired` successor), stranding the outgoing session un-expired. That inner lock is Path A of an AB-BA cycle whose opposing `self`→`currentSessionMutex` edge is removed in **#174 (C-968)** — coordinated so the two PRs stay coherent and C-967 stays out of the cycle.

## Guards

This race class has no reliable in-process crash signal (`RACE_TESTING.md` §5), so the guards are structural and deterministic (`RaceTripwireTests`): idempotent removal, replacement deterministically detaching the outgoing session, plus a green-stable concurrent-detach check. Revert-check ran isolated on the test lane — green with the fix (3/3), red with it reverted (2/2: idempotence → `NSRangeException: not registered as an observer`; replacement → outgoing not flagged).

Not claiming this clears a specific production crash bucket — it makes the observer lifecycle correct, bounded, and verifiable; bucket movement is confirmed separately post-release.

---

Linear: https://linear.app/teakio/issue/C-967

### DIFF
--- a/Automated/AutomatedTests/RaceTripwireTests.m
+++ b/Automated/AutomatedTests/RaceTripwireTests.m
@@ -617,13 +617,14 @@ static NSMutableArray* keepAliveBareSessions;
 // --- Cross-session deviceConfiguration KVO observer-lifecycle guards ---
 //
 // Every session registers KVO observers on the process-shared deviceConfiguration in -init, and
-// those must be removed exactly once. The fix removes them deterministically at session replacement
-// under currentSessionMutex — the same lock every -init addObserver holds — so all add/remove on the
-// shared object are serialized, and makes the removal idempotent so -dealloc's fallback can't
-// double-remove. The cross-thread add/remove race itself has no reliable in-process crash signal
-// (RACE_TESTING.md §5), so these two guards pin the structural invariants deterministically (clean
-// assertions, not a crash repro): idempotent removal, and the removal actually happening at
-// replacement rather than being left to a background dealloc.
+// those must be removed exactly once. The fix serializes every add and remove on the shared object
+// under one dedicated leaf lock (deviceConfigurationObserverMutex, held directly in -init and in
+// detach), removes them deterministically at session replacement rather than from a background
+// -dealloc, and makes the removal idempotent so -dealloc's fallback can't double-remove. The
+// cross-thread add/remove race itself has no reliable in-process crash signal (RACE_TESTING.md §5),
+// so these two guards pin the structural invariants deterministically (clean assertions, not a crash
+// repro): idempotent removal, and the removal actually happening at replacement rather than being
+// left to a background dealloc.
 
 // Idempotent removal. -init registered the observers, so the first detach removes them and the
 // second must no-op; without the flag guard the second removeObserver throws "not registered as an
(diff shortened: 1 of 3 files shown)


AlexSc (changes requested):
I don't know about this one. This feels very "works now" but with distributed dynamics that require a careful read. We have one lock to ensure only one session at a time can unregister, and a different lock that happens to function as a registration lock, and strict ordering requirements at several call sites to make sure it all works.

I would feel better with a single lock around all KVO operations so serialization is verifiable with a trivial read of the code.

I'm also not clear if removing the currentSession lock is a good idea.

I'm open to push back on this if I'm misunderstanding something. I think it would also be good to do an online docs search to verify KVO registration is not thread safe.

AlexSc (approved):

PR #179, opened 2026-07-07
Serialize TeakUserProfile's residual -send dict read on the TeakRequest retry ladder

TeakBuilds:
Linear: https://linear.app/teakio/issue/C-959/teakuserprofile-serialize-residual-send-dict-read-on-teakrequest-retry

## What
C-955 (#163) moved `stringAttributes`/`numberAttributes` writes onto the serial `io.teak.sdk.operationQueue`, but `-send`'s own read of those dicts still ran on whatever queue invoked it. `TeakRequest`'s retry ladder (socket-error and server-retry branches, `TeakRequest.m:393`/`405`) re-invokes `[self send]` from `dispatch_get_main_queue()` after a delay — if a new attribute set call re-arms `scheduledBlock` in that window, `TeakUserProfile`'s `-send` override runs on the main queue and copies the same dicts the operationQueue writer can be mutating concurrently. Same mutable-container UAF class C-955 closed, reopened at a different entry point into `-send`.

## Fix
Wrap `-send`'s whole body in `dispatch_async([Teak operationQueue], ...)`, mirroring C-955's fix for the setter. Every call path (the batching `scheduledBlock`, `TeakSession`'s flush-on-Expiring, and the retry ladder) now serializes on `operationQueue` regardless of which queue invoked `-send`. All three callers are fire-and-forget — verified none depend on synchronous completion or ordering.

## Test
Adapted `testUserProfileAttributeDictIsSerializedUnderConcurrency`'s TSan template in `RaceTripwireTests.m`: one thread drives the real `setStringAttribute` (write, on operationQueue), the other calls `-send` directly off-queue (blackholed), standing in for the retry ladder's off-queue re-invocation. Revert-checked: reverting the `dispatch_async` hop reproduces `ThreadSanitizer: race on NSMutableDictionary` between the setter's write and `-send`'s read; restoring the fix is silent. Full regular suite also green (162/162).

## Changelog
No `4.3.14.yaml` entry per the current wave's protocol — changelog line below for whoever assembles it:

> Fixed a residual thread-safety issue in user profile attribute sending that could occur during request retries.

## Note for release scoping
This ticket's own body says "post-4.3.14 monitor-and-address" — it's included in this wave to complete the C-955 stack, not because it's in an observed crash family yet. Flagging as a candidate to defer out of 4.3.14 if desired.

### DIFF
--- a/Teak/Core/TeakUserProfile.m
+++ b/Teak/Core/TeakUserProfile.m
@@ -65,24 +65,31 @@
 }
 
 - (void)send {
-  // No scheduledBlock means no pending update
-  if (self.scheduledBlock != nil) {
-    dispatch_block_cancel(self.scheduledBlock);
-    self.scheduledBlock = nil;
+  // TeakRequest's retry ladder can re-invoke -send from dispatch_get_main_queue() after a
+  // delay (socket-error and server-retry branches); if a new attribute has re-armed
+  // scheduledBlock in that window, this would read stringAttributes/numberAttributes off
+  // the serial operationQueue that setAttribute:forKey:inDictionary: mutates them on. Hop
+  // onto that same queue so this read is never concurrent with that write.
+  dispatch_async([Teak operationQueue], ^{
+    // No scheduledBlock means no pending update
+    if (self.scheduledBlock != nil) {
+      dispatch_block_cancel(self.scheduledBlock);
+      self.scheduledBlock = nil;
 
-    NSMutableDictionary* payload = [self.payload mutableCopy];
-    [payload addEntriesFromDictionary:@{
-      @"string_attributes" : [self.stringAttributes copy],
-      @"number_attributes" : [self.numberAttributes copy],
-      @"context" : [self.context copy],
-      @"ms_since_first_event" : [NSNumber numberWithDouble:[self.firstSetTime timeIntervalSinceNow] * -1000.0]
-    }];
-    self.payload = payload;
+      NSMutableDictionary* payload = [self.payload mutableCopy];
+      [payload addEntriesFromDictionary:@{
+        @"string_attributes" : [self.stringAttributes copy],
+        @"number_attributes" : [self.numberAttributes copy],
+        @"context" : [self.context copy],
+        @"ms_since_first_event" : [NSNumber numberWithDouble:[self.firstSetTime timeIntervalSinceNow] * -1000.0]
+      }];
+      self.payload = payload;
 
-    [super send];
+      [super send];
 
-    self.firstSetTime = nil;
-  }
+      self.firstSetTime = nil;
+    }
+  });
 }
 
 @end
--- a/Automated/AutomatedTests/RaceTripwireTests.m
+++ b/Automated/AutomatedTests/RaceTripwireTests.m
@@ -73,6 +73,14 @@ extern NSString* const currentSessionMutex;
 
 @interface TeakUserProfile (RaceTripwire)
 @property (strong, nonatomic) NSMutableDictionary* stringAttributes;
+@property (strong, nonatomic) NSMutableDictionary* numberAttributes;
+@property (strong, nonatomic) NSString* context;
+@end
+
+// blackhole is readonly in the public header, readwrite in TeakRequest+Internal.h; re-exposed here
+// per the project convention so the tripwire can suppress the real network send.
+@interface TeakRequest (RaceTripwire)
+@property (nonatomic, readwrite) BOOL blackhole;
 @end
 
 // countryCode/serverSessionId/facebookAccessToken are private (declared only in TeakSession.m's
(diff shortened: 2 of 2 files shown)


Teak/Core/TeakUserProfile.m
```diff
@@ -65,24 +65,31 @@ - (void)setAttribute:(id)value forKey:(NSString*)key inDictionary:(NSMutableDict
...
+    if (self.scheduledBlock != nil) {
+      dispatch_block_cancel(self.scheduledBlock);
+      self.scheduledBlock = nil;
+
+      NSMutableDictionary* payload = [self.payload mutableCopy];
+      [payload addEntriesFromDictionary:@{
+        @"string_attributes" : [self.stringAttributes copy],
+        @"number_attributes" : [self.numberAttributes copy],
+        @"context" : [self.context copy],
+        @"ms_since_first_event" : [NSNumber numberWithDouble:[self.firstSetTime timeIntervalSinceNow] * -1000.0]
```
AlexSc, line 84: If we retry, but there hasn't been a subsequent attribute set, firstSetTime will be nil, so we'll ultimately call numberWithDouble:nil. Is that safe?

TeakBuilds: [sdks-worker-c-959] It's unreachable, not just ABI-safe: `-send`'s whole body — including this line — only runs when `scheduledBlock != nil` (line 75). `scheduledBlock` only ever goes nil→non-nil inside `setAttribute:forKey:inDictionary:`, which unconditionally stamps `firstSetTime` (if nil) *before* arming `scheduledBlock`, in the same serialized operationQueue block (TeakUserProfile.m:44-57 on this branch). So by the time any `-send` invocation observes `scheduledBlock` as non-nil, `firstSetTime` is already non-nil too — they're set atomically together, and post-C-959 both are only ever read on that same serial queue.

Concretely for the retry case: if there's no subsequent attribute set, `scheduledBlock` is still nil (the first `-send` nil'd it before `[super send]`), so the guard fails and we never reach this line at all — the retry is a silent no-op (the reviewer's fyi, pre-existing, not this PR's concern). If a new attribute *did* re-arm `scheduledBlock` during the retry delay, that same setAttribute call already stamped `firstSetTime`, so it's non-nil when we get here.

Belt-and-suspenders: even if it were somehow reached with a nil `firstSetTime`, message-to-nil for a `double`-returning method is well-defined as 0.0 on the arm64/x86_64 targets this SDK ships for — not a crash. But the real answer is the guard makes that case unreachable.

TeakBuilds: [sdks-worker-c-959] Tightening the above — "set atomically together" undersold it since firstSetTime isn't set-once: `-send` clears it at line 90, right after using it at 84. The real invariant is **scheduledBlock-non-nil ⟹ firstSetTime-non-nil**, and it holds on both write paths:

- `setAttribute:forKey:inDictionary:` stamps `firstSetTime` (44) strictly before arming `scheduledBlock` (57)
- `-send` nils `scheduledBlock` (77) strictly before clearing `firstSetTime` (90) — both inside the same `scheduledBlock != nil` guard, so there's no window where `scheduledBlock` is non-nil but `firstSetTime` is nil

What guarantees no third call tears that pairing (e.g. a `setAttribute` landing between 77 and 90 to re-arm `scheduledBlock` while `firstSetTime` is about to be cleared) is this PR's own fix — both blocks now run serialized on `operationQueue`, so line 90's clear and any concurrent `setAttribute`'s stamp-then-arm can never interleave. So it's not just that line 84 happens to be unreachable — this PR's serialization is what makes that guarantee hold.

AlexSc: So in practice if there's a network hiccup and this request doesn't go out the request retry logic doesn't actually retry the request?

TeakBuilds: [sdks-worker-c-959] Confirmed, yes — traced it through: on the first attempt, `-send` nils `scheduledBlock` (77) and calls `[super send]` (88) *before* it knows whether that attempt succeeds. `TeakRequest`'s retry ladder (`TeakRequest.m:393-394` socket-error path, `:405-406` 5xx/server-retry path) re-invokes `[self send]` after a delay — polymorphic dispatch lands that back on `TeakUserProfile`'s override, not some independent request-layer repost (confirmed: the base `TeakRequest -send` at `TeakRequest.m:312` always builds a fresh `NSMutableURLRequest` from `self.payload` at call time, there's no cached-request resend path). But by the time that delayed call fires, `scheduledBlock` is already nil and nothing re-armed it, so the guard at line 75 fails — the retry is a silent no-op, no new HTTP request goes out.

This is fully pre-existing — I diffed against `origin/4.3-stable` and the guard/nil-before-send structure is byte-identical there. C-959 only changes which queue reads it; it doesn't touch this logic at all.

On data loss: `stringAttributes`/`numberAttributes` are never cleared in `-send` (only `scheduledBlock` and `firstSetTime` are), so the values aren't erased — they're deferred, not dropped, *as long as another `setAttribute` call happens before the session ends*: any later call re-arms `scheduledBlock` and the next successful `-send` re-serializes the full current dicts, so the stale values ride along. But if no further attribute change happens in that session, the update never lands — the next session's `TeakUserProfile` is reseeded from the server's `user_profile` dict (`TeakSession.m:319`), which never saw the failed update, so it's genuinely gone at that point, not just delayed.

So: real gap, but in the retry *semantics*, not the race this PR fixes. Flagging to the lead as a followup candidate rather than folding into #179.

AlexSc (approved):

PR #157, opened 2026-06-30
Document the personalizationData notification scheduling API

TeakBuilds:
## Summary

`scheduleNotificationForCreative:secondsFromNow:personalizationData:` (C: `TeakNotificationSchedulePersonalizationData`) shipped in 4.3.0 as the replacement for the deprecated `withMessage:` variant, but `working-with-teak.adoc` never got updated — it only documented the deprecated old API.

Updates the "Scheduling a Notification" section to:
- Lead with the new personalizationData API and its C signature
- Show how to read the `TeakOperation` result (async/poll-style, unlike the old synchronous `TeakNotification` return) — the example is sourced from `Sample/ViewController.m`'s real, already-exercised `scheduleNotification:` code, not an invented loop
- Demote the old `withMessage:`/`TeakNotificationSchedule` call to a clearly marked deprecated callout
- Fix an unrelated "do be delivered" typo in the adjacent long-distance scheduling paragraph

Docs-only change, no code touched, no changelog entry (not a behavior change).

Linear: https://linear.app/teakio/issue/C-385

## Test plan
- [x] Hand-checked AsciiDoc syntax (admonition format, backtick pairing, xref anchors) against existing conventions in this file and the wider docs corpus
- [ ] Reviewer pass

PR #189, opened 2026-07-22
Document scheduleNotificationForCreative:secondsFromNow:forUserIds: (cross-user scheduling)

TeakBuilds:
## Summary

`scheduleNotificationForCreative:secondsFromNow:forUserIds:` is a live, non-deprecated 4.3.x API on `TeakNotification` that schedules a creative for other players (by game-assigned user id), but it wasn't in the shipped docs -- it was drafted once in #157 and deliberately cut to keep that PR scoped to the personalizationData variant.

Adds a "Scheduling a Notification for Other Players" subsection to `working-with-teak.adoc`, matching the conventions that section settled on after the #157 cut (paired Objective-C/Swift examples, no C bridge signatures, no defensive framing). Both example snippets were compiled (`clang -fsyntax-only` / `swiftc -typecheck`) against the real prebuilt `Teak.xcframework`, not hand-derived.

Notes the behavioral gotchas, both confirmed against source (not the header comments):
- The returned `TeakNotification` completes asynchronously -- check `.completed` before `.status`.
- The creative must already exist on the dashboard -- confirmed by @alex: only the deprecated `withMessage:` variants auto-create one.
- `.teakNotifId` is a JSON-encoded array of the delivery ids that were created, not a single id, and it's **not** positionally matched to `userIds` -- confirmed against `BulkNotificationScheduleController#schedule_from_device` in carrot: the array comes back in ApiKey query order, and any user id with no matching player in-game is silently dropped, so it can be shorter than the input.

Docs-only change, no code touched, no changelog entry (not a behavior change).

A separate follow-up (C-1167) was filed for an unrelated, pre-existing header-comment inaccuracy this PR's discussion surfaced (`TeakNotification.h:14`, the personalizationData variant) -- out of scope here, doesn't touch this diff.

Linear: https://linear.app/teakio/issue/935

## Test plan
- [x] Both example snippets compiled clean against `../teak-ios-framework`'s `Teak.xcframework` (ObjC + Swift)
- [x] Xref target (`scheduleNotificationForCreative:secondsFromNow:forUserIds:`) verified against fresh `npm run docs` doxygen output -- exact anchor match
- [x] Both deferred server-behavior questions confirmed and folded into the prose
- [x] Reviewer pass

### DIFF
--- a/docs/modules/ROOT/pages/working-with-teak.adoc
+++ b/docs/modules/ROOT/pages/working-with-teak.adoc
@@ -257,7 +257,7 @@ let notif = TeakNotification.scheduleNotification(forCreative: "my_creative_id",
                                                     forUserIds: ["user1", "user2"])
 ----
 
-This returns a ``TeakNotification`` that completes asynchronously -- check ``notif.completed`` before reading ``notif.status``. ``notif.teakNotifId`` is a JSON-encoded array of scheduled-notification ids rather than a single id.
+This returns a ``TeakNotification`` that completes asynchronously -- check ``notif.completed`` before reading ``notif.status``. ``notif.teakNotifId`` is a JSON-encoded array of the delivery ids that were created, not a single id -- it isn't positionally matched to ``userIds`` and may contain fewer entries, for example if one of the user ids you passed doesn't resolve to a player in your game.
 
 === Canceling a Scheduled Notification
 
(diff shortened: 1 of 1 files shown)


AlexSc (approved):
Also make sure the commits from this branch land on 4.3-stable so we don't lose track of what's been documented while working on patch releases

PR #196, opened 2026-07-26
Fix teakNotifId header doc and mistagged JSON-serialization error

TeakBuilds:
https://linear.app/teakio/issue/C-1235/teak-ios-teaknotifid-header-doc-says-the-identifier-its-a-json-array
https://linear.app/teakio/issue/C-1236/teak-ios-schedule-path-logs-json-failure-under-the-cancel-all-tag

[teak-teak-ios-worker-c-1235]

Carries C-1236 too — same method, ten lines apart, not worth splitting.

- `TeakNotification.h` — `teakNotifId`'s doc said "The identifier for the scheduled notification" (singular). True for the single-user schedule/cancel paths, but `forUserIds:` and `cancelAll` both JSON-encode an array of ids into it. Reworded to one sentence covering both shapes instead of enumerating each writer (avoids the doc going stale again if a fifth writer shows up).
- `TeakNotification.m:234` — the `forUserIds:` JSON-serialization-failure log was tagged `notification.cancel_all.error.json`, copy-pasted from the `cancelAll` path. Retagged to `notification.schedule.error.json`, mirroring the existing `cancel_all.error`/`cancel_all.error.json` split so a serialization failure stays greppable apart from a server-rejected schedule. (Deviates from the issue's literal `notification.schedule.error` suggestion — confirmed with lead-pm.)

Doc-only + log-tag-only, no behavior change.

Checked teak-android/teak-unity for the same singular-doc defect: no match. Android's `teakNotifId` is an unrelated `long` (received-push platform id, not a schedule-operation result). Unity's equivalent (`Reply.Notifications`, a `List<Notification>`) already documents itself plural — it's already parsing this same iOS property as "sometimes an array" (`TeakNotification.cs:457-492`), which is corroborating evidence the iOS header was the thing that was wrong.

### DIFF
--- a/Teak/TeakNotification.h
+++ b/Teak/TeakNotification.h
@@ -18,9 +18,9 @@
 + (nullable TeakOperation*)scheduleNotificationForCreative:(nonnull NSString*)creativeId secondsFromNow:(int64_t)delay personalizationData:(nullable NSDictionary*)personalizationData;
 
 /**
- * The identifier for the scheduled notification.
+ * The id of the notification this call created or canceled, or a JSON-encoded array of ids if it affected more than one.
  *
- * Also accessable via:
+ * Also accessible via:
  *
  * 	const char* TeakNotificationGetTeakNotifId(TeakNotification* notif)
  */
@@ -37,7 +37,7 @@
  * - error.parameter.delayInSeconds - delayInSeconds can not be negative, or greater than one month
  * - error.parameter.userIds - userIds can not be null or empty
  *
- * Also accessable via:
+ * Also accessible via:
  *
  * 	const char* TeakNotificationGetStatus(TeakNotification* notif)
  */
@@ -96,7 +96,7 @@
 /**
  * YES if the notification operation has completed.
  *
- * Also accessable via:
+ * Also accessible via:
  *
  * 	BOOL TeakNotificationIsCompleted(TeakNotification* notif)
  */
--- a/Teak/TeakNotification.m
+++ b/Teak/TeakNotification.m
@@ -231,7 +231,7 @@
                                                       NSError* error = nil;
                                                       NSData* jsonData = [NSJSONSerialization dataWithJSONObject:reply[@"ids"] options:0 error:&error];
                                                       if (error) {
-                                                        TeakLog_e(@"notification.cancel_all.error.json", @{@"value" : reply[@"ids"], @"error" : error});
+                                                        TeakLog_e(@"notification.schedule.error.json", @{@"value" : reply[@"ids"], @"error" : error});
                                                         ret.teakNotifId = @"[]";
                                                       } else {
                                                         ret.teakNotifId = [[NSString alloc] initWithData:jsonData encoding:NSUTF8StringEncoding];


TeakBuilds (commented):
[teak-teak-ios-reviewer-c-1235] **CHANGES REQUIRED**

Diff itself is correct — both lines. Retag matches the `cancel_all.error`/`cancel_all.error.json` split, `TeakLog_e(NSString*, NSDictionary*)` is a real overload so it builds, and the header claim holds against all four `teakNotifId` writers (`:170` single, `:235/237` array, `:278` single, `:309/311` array). No contradiction left with `working-with-teak.adoc:260`. The one blocking item is in the commit message, not the code.

**1. Commit message still says Unity has no equivalent property — `major`, high confidence**

The PR body got corrected; the commit message didn't. "Key decisions made" bullet 3 still reads:

> unity has no such property (just inline marshaling of the iOS C bridge call)

That's the false negative, and git history outlives the PR body. Unity has two equivalent surfaces — `TeakNotification.Reply.Notifications` (`List<Notification>`) and `Teak.Notification.Reply.ScheduleIds` (`List<string>`) — both already documented plural. Amend to match the corrected body.

Two things worth fixing in the same amend:

- The bullet records the *conclusion* but not *why the first grep missed it* — it searched the literal `teakNotifId` against a C# codebase that names the concept `ScheduleId`. That's the reusable part.
- Bullet 1 credits the cancelAll roll-in to lead-pm; it came from plan review and lead-pm endorsed it. Bullet 2 already says "per reviewer's reasoning" — worth the same precision, since the log's value is tracing who caught what.

**2. `accessable` → `accessible` at `TeakNotification.h:23` — `minor`, high confidence**

Inside the doc comment block this PR is editing, two lines under the changed line, rendering into the same generated API paragraph the issue is about. Free while you're amending. `:40` and `:99` in this file are the same typo — take all three or just `:23`, your call. The other 8 live in `Teak.h` and are a separate sweep, not this PR's job.

**3. "single-notification operation" / "multi-notification operation" — `fyi`, take it or leave it**

Not relitigating "JSON-encoded array" vs "a List" — you and lead-pm are right that the deserialize obligation is load-bearing for an ObjC reader, and I wasn't proposing to drop it. Different point: the two *operation* terms aren't defined anywhere in the header, so the reader has to work out which bucket `cancelAll` falls in. Something like "The id of the notification this call created or canceled, or a JSON-encoded array of ids if it affected more than one" keeps the JSON-encoding explicit, drops the undefined terms, and says *which* ids without duplicating the guide's per-call detail. Your wording is accurate as-is — this is polish, not a correction.

TeakBuilds (commented):
[teak-teak-ios-reviewer-c-1235] **APPROVED**

All three findings addressed. Re-read the cumulative diff against `4.3-stable`: header wording, three `accessible` fixes in this file, log retag. Nothing else moved.

**On the append-only call — you're right, and my finding over-specified.** I asked for an amend; what I actually needed was that the durable record not assert a falsehood. A correction commit clears that bar by a different route, and it has a virtue the amend didn't: `CLAUDE.md` says the interaction log exists as an audit log of how we work with Claude, and an audit log that silently deletes a wrong conclusion loses the thing it's for. 3f31700 records the claim, the correction, *and* the root cause — search by concept, not by the other language's symbol name. That's the version worth keeping. It also survives every merge shape here: squash concatenates both messages, a merge commit keeps both. Either way the correction travels with the claim.

Confirmed independently rather than taking the plan's word for it: nothing under `Automated/` references the old tag string or `teakNotifId`, so the retag can't break a test. `build` is green; `test`/`test_race` still pending at time of writing — that's the APM's gate, not a code concern.

Header now reads accurate against all four writers, keeps the deserialize obligation explicit, and no longer leans on terms the header never defines. Comment is timeless — no ticket refs, no roadmap language.

One candidate follow-up, **not** for this PR: the same `accessable` typo is in `Teak.h` 8 more times. This PR correctly scoped to the file it was already editing, but that leaves the two headers inconsistent. Raising it in-channel for lead-pm to disposition — a one-line sed sweep, whenever someone's next in `Teak.h`.

AlexSc (approved):

